PR #8, opened 2018-04-21
updated strategies to work with latest freqtrade version

berlinguyinca:
please merge as see fit.


user_data/strategies/strategy002.py
```diff
@@ -16,7 +16,7 @@
 class_name = 'CustomStrategy'
```
gcarq: `class_name` is no longer necessary and can be removed

berlinguyinca: oh I thought I did this on most files. My bad slipped by

glonlas:
Thank you very much for the PR.

PR #24, opened 2018-12-27
Update free trading strategies table on README.md

oonid:
## Summary

Change reference of free trading strategies table to files in official repo.
Avoid confusion from old issues that still refer to an external (unofficial) repo.


xmatthias:
:+1: superb, thanks a lot!

PR #35, opened 2019-07-11
MACDStrategy_crossed

hroff-1902:
I'm not sure if it's proper place for yet another strategy since these are the strategies by @berlinguyinca as the name of the catalog says, hope it fits...

This is a strategy complement to MACDStrategy: it uses crossed_above() and crossed_below() instead of comparison of macd with macdsignal.

I found it would be interesting and probably useful to have such a couple of complement strategies here 1) for testing purposes, 2) to have an ability to compare the behaviour of these 2 similar signals at different markets and market conditions.

Minor cosmetics to original MACDStrategy included.

### DIFF
--- a/user_data/strategies/berlinguyinca/MACDStrategy_crossed.py
+++ b/user_data/strategies/berlinguyinca/MACDStrategy_crossed.py
@@ -1,0 +1,77 @@
+
+# --- Do not remove these libs ---
+from freqtrade.strategy.interface import IStrategy
+from typing import Dict, List
+from functools import reduce
+from pandas import DataFrame
+# --------------------------------
+
+import talib.abstract as ta
+import freqtrade.vendor.qtpylib.indicators as qtpylib
+
+
+class MACDStrategy_crossed(IStrategy):
+    """
+        buy:
+            MACD crosses MACD signal above
+            and CCI < -50
+        sell:
+            MACD crosses MACD signal below
+            and CCI > 100
+    """
+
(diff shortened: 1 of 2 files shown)


user_data/strategies/berlinguyinca/MACDStrategy_crossed.py
```diff
@@ -0,0 +1,77 @@
...
+    def populate_buy_trend(self, dataframe: DataFrame) -> DataFrame:
+        """
+        Based on TA indicators, populates the buy signal for the given dataframe
+        :param dataframe: DataFrame
+        :return: DataFrame with buy column
+        """
+        dataframe.loc[
+            (
+                qtpylib.crossed_above(dataframe['macd'], dataframe['macdsignal']) &
+                (dataframe['cci'] <= -50.0)
```
xmatthias: I think we should always have a check for `dataframe['volume'] > 0`.

This was identified as a problem a while ago in the main repo - and not having it can produce odd buys which you cannot archive in reality since you don't have buyers/sellers show up.

I know we don't have that for most strategies (but we should) 

Apart from that it's great :+1:

hroff-1902: I have not found this check (`dataframe['volume'] > 0`) in any strategy in freqtrade-strategies.

Let's make it in another PR, as a complex change. This strategy simply complements the MACDStrategy, differs in that it uses crosses instead of comparison...

Issue #33, opened 2019-07-07
hyperopts for the strategies


hroff-1902:
All the strategies published in freqtrade-strategies should be supplied with corresponding Hyperopts to adjust the constants used in them basing on actual historical data.

For example:
There should be MACDHyperopt for the MACDStrategy (https://github.com/freqtrade/freqtrade-strategies/blob/master/user_data/strategies/berlinguyinca/MACDStrategy.py) which would allow to adjust upper and lower borders for CCI band, which is now hardcoded to (-50.0, 100.0), as well as to select best roi and stoploss.

berlinguyinca:
great suggestions

OtenMoten:
Hello dear community,

I wrote the hyper-optimization for the `AverageStrategy.py` strategy from [github.com/freqtrade/freqtrade-strategies/blob/master/user_data/strategies/berlinguyinca/](https://github.com/freqtrade/freqtrade-strategies/blob/master/user_data/strategies/berlinguyinca/AverageStrategy.py) written by @berlinguyinca 

I've created a pull-request and when it's reviewd I will link it here. <br>
[ >>> PLACEHOLDER <<<]

It seems to be a nice idea to gather all the hyper-optimization related to the strategies from @berlinguyinca here in this issue.

I wondered about the range of the moving average, therefore I created a hyper-optimization that calculates the sweet-spot for buy and sell when short-ma and medium-ma cross above and vice versa. I.E.: @berlinguyinca used in his version of `AverageStrategy.py` the timeframes for `ma-short = 8` and `ma-medium = 21`. I want to know what are the best timeframes beside `8` and `21`. So, the algorithm is looping through all the different timeframes of the both moving averages.

Hopefully you get value from this.

Kind regards
Kevin

gerardsimons:
Hey there, great work there. I am facing some issues though. I am trying to run your HyperOpt class lik so :

`freqtrade hyperopt --config config.json --hyperopt AverageHyperopt --hyperopt-loss ShortTradeDurHyperOptLoss --strategy AverageStrategy -e 500 `

but keep getting this error : 

```
Traceback (most recent call last):
  File "/home/gerard/Git/freqtrade/freqtrade/main.py", line 37, in main
    return_code = args['func'](args)
  File "/home/gerard/Git/freqtrade/freqtrade/commands/optimize_commands.py", line 83, in start_hyperopt
    hyperopt.start()
[...]

PR #36, opened 2019-07-11
hyperopt for MACDStrategy

hroff-1902:
Here it is, as it was proposed in #33, an implementation (as a beginning and inspiration for other authors) of the hyperopt for MACDStrategy.

Created from the generic hyperopt template, it allows to optimize the CCI band boundaries in the MACDStrategy strategy.

Thanks to @Prossi79, this is partly his work.

### DIFF
--- a/user_data/hyperopts/MACDStrategy_hyperopt.py
+++ b/user_data/hyperopts/MACDStrategy_hyperopt.py
@@ -3,12 +3,11 @@
 import talib.abstract as ta
 from pandas import DataFrame
 from typing import Dict, Any, Callable, List
-from functools import reduce
 
-import numpy
+# import numpy as np
 from skopt.space import Categorical, Dimension, Integer, Real
 
-import freqtrade.vendor.qtpylib.indicators as qtpylib
+# import freqtrade.vendor.qtpylib.indicators as qtpylib
 from freqtrade.optimize.hyperopt_interface import IHyperOpt
 
 class_name = 'MACDStrategy_hyperopt'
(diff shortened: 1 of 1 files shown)


user_data/hyperopts/MACDStrategy_hyperopt.py
```diff
@@ -0,0 +1,164 @@
...
+class MACDStrategy_hyperopt(IHyperOpt):
+    """
+    This is a test hyperopt to inspire you.
+    More information in https://github.com/freqtrade/freqtrade/blob/develop/docs/hyperopt.md
+     You can:
+    - Rename the class name (Do not forget to update class_name)
+    - Add any methods you want to build your hyperopt
+    - Add any lib you need to build your hyperopt
+     You must keep:
+    - the prototype for the methods: populate_indicators, indicator_space, buy_strategy_generator,
```
xmatthias: I don't wanna sound picky - but i think we should replace most of this doc-string with an explanation on what is optimized (this hyperopt file assumes you know you wanna use MACD crossover with CCI protection) - and you only optimize CCI. That's a valid usecase, and it's explained in the PR description - but that's not visible to the regular user ...

Example (from your comment):

> it allows to optimize the CCI band boundaries in the MACDStrategy strategy.

hroff-1902: agree, will change the doc strings

xmatthias: i guess you forgot about this just as i did ... ?

hroff-1902: A bit... ))

PR #39, opened 2019-08-10
update interfaces for populate_* methods

hroff-1902:
Interfaces to populate_methods refreshed.

* added `metadata: dict` parameter


hroff-1902:
> With this think we can think of removing the "legacy" method from the strategy (which is something i remembered a week or so ago) ... at least once we've given this change a week or two

Exactly! I also decided that it's time to propose the deletion of `advise_*()` methods.

xmatthias:
no i would leave the advise abstraction to be honest - it'll allow us to keep the log we have now - AND simplify  another interface change should we need one

hroff-1902:
> no i would leave the advise abstraction to be honest - it'll allow us to keep the log we have now - AND simplify another interface change should we need one

Okay.

PR #43, opened 2019-08-21
CombinedBinHAndCluc code clean

il-katta:
removed unused indicators from strategy CombinedBinHAndCluc

### DIFF
--- a/user_data/strategies/berlinguyinca/CombinedBinHAndCluc.py
+++ b/user_data/strategies/berlinguyinca/CombinedBinHAndCluc.py
@@ -1,17 +1,17 @@
 # --- Do not remove these libs ---
+import freqtrade.vendor.qtpylib.indicators as qtpylib
 import numpy as np
 # --------------------------------
 import talib.abstract as ta
+from freqtrade.strategy.interface import IStrategy
 from pandas import DataFrame
 
-import freqtrade.vendor.qtpylib.indicators as qtpylib
-from freqtrade.strategy.interface import IStrategy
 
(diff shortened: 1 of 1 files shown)


user_data/strategies/berlinguyinca/CombinedBinHAndCluc.py
```diff
@@ -1,17 +1,17 @@
...
 
-import freqtrade.vendor.qtpylib.indicators as qtpylib
-from freqtrade.strategy.interface import IStrategy
 
 def bollinger_bands(stock_price, window_size, num_of_std):
     rolling_mean = stock_price.rolling(window=window_size).mean()
     rolling_std = stock_price.rolling(window=window_size).std()
     lower_band = rolling_mean - (rolling_std * num_of_std)
-    return rolling_mean, lower_band
+    return np.nan_to_num(rolling_mean), np.nan_to_num(lower_band)
```
xmatthias: I think this is dangerous (mainly during backtesting), since bollinger bands (like almost every other indicator) requries some "startup-time" (`window_size`) at the beginning of the time period.

Using `np.nan_to_num()` removes this and fakes values by using 0, which (depending on the conditions) may trigger buys which would not happen in reality.

il-katta: I didn't added, just move from `populate_indicators` function.
However in `populate_buy_trend` function first condition are
`dataframe['lower'].shift().gt(0)` so fake buy may not appended.

PR #47, opened 2019-10-19
The strategy should use rolling volume, not absolute mean

xmatthias:
## Summary
Since these serve as samples, they should not contain logical errors.

using `dataframe['volume'].mean()` produces completely different signals during dry-run and backtesting - especially when backtesting longer periods.

Using a rolling mean instead will correct this. 200 was selected arbitrarily - however yields better results for  20190101-20190930 than using the mean of the whole dataframe.


xmatthias:
ta_on_candle was the initial proposal - but was renamed to process_only_new_candle before merging (iirc)... 
so `ta_on_candle` never had any "productive" use...

PR #73, opened 2020-04-04
stochf parameters are all integers

xmatthias:
## Summary
The latest version of ta-lib (not released, master) enforces parameters to be integers.

We're not always following this (especially with STOCHF) .


hroff-1902:
hmm, do you have a link into the issue/news message for ta-lib for this?

Issue #68, opened 2020-02-09
TD Sequential indicator  - new strategy


Prossi79:
*For requestion a new strategy. Please use the template below.*  
*Any strategy request that does not follow the template will be closed.*

## Step 1: What indicators are required?
no indicators needed, just close prices

## Step 2: Explain the Buy Strategy

the trigger to buy -> 
When you see 9 consecutive closes “lower” than the close 4 bars prior.
An ideal buy is when the low of bars 6 and 7 in the count are exceeded by the low of bars 8 or 9.

## Step 3: Explain the Sell Strategy
trigger to sell ->
When you see 9 consecutive closes “higher” than the close 4 candles prior.
An ideal sell is when the the high of bars 6 and 7 in the count are exceeded by the high of bars 8 or 9.

## Source
What come from this strategy? Cite your source:
https://hackernoon.com/how-to-buy-sell-cryptocurrency-with-number-indicator-td-sequential-5af46f0ebce1

hroff-1902:
* a good article on this indicator: http://practicaltechnicalanalysis.blogspot.com/2013/01/tom-demark-sequential.html
* a reference page at the tradingview: https://www.tradingview.com/chart/BTCUSD/2MJxxId6-Tom-Demark-Sequential-T-D-Sequential/

(to log it here)

xmatthias:
I guess #75 does fix this?

Let me know if that's not the case ...

PR #75, opened 2020-04-17
TD sequential strategy

bmoulkaf:
solving issue #68

### DIFF
--- a/user_data/strategies/berlinguyinca/TDSequentialStrategy.py
+++ b/user_data/strategies/berlinguyinca/TDSequentialStrategy.py
@@ -28,9 +28,20 @@ class TDSequentialStrategy(IStrategy):
 
     # Optimal stoploss designed for the strategy
     stoploss = -0.05
+    
+    # Trailing stoploss
+    trailing_stop = False
+    # trailing_only_offset_is_reached = False
+    # trailing_stop_positive = 0.01
+    # trailing_stop_positive_offset = 0.0  # Disabled / not configured
 
     # Optimal ticker interval for the strategy
     ticker_interval = '1h'
+    
+    # These values can be overridden in the "ask_strategy" section in the config.
+    use_sell_signal = True
+    sell_profit_only = False
+    ignore_roi_if_buy_signal = False
 
     # Optional order type mapping
     order_types = {
--- a/user_data/strategies/berlinguyinca/TDSequentialStrategy.py
+++ b/user_data/strategies/berlinguyinca/TDSequentialStrategy.py
@@ -8,6 +8,8 @@ from freqtrade.strategy.interface import IStrategy
 class TDSequentialStrategy(IStrategy):
     """
     Strategy based on TD Sequential indicator.
+    source:
+    https://hackernoon.com/how-to-buy-sell-cryptocurrency-with-number-indicator-td-sequential-5af46f0ebce1
     
     Buy trigger:
         When you see 9 consecutive closes "lower" than the close 4 bars prior.
--- a/user_data/strategies/berlinguyinca/TDSequentialStrategy.py
+++ b/user_data/strategies/berlinguyinca/TDSequentialStrategy.py
@@ -1,0 +1,132 @@
+import talib.abstract as ta
+from pandas import DataFrame
+import scipy.signal
+import freqtrade.vendor.qtpylib.indicators as qtpylib
+from freqtrade.strategy.interface import IStrategy
+
+
+class TDSequentialStrategy(IStrategy):
+    """
+    Default Strategy provided by freqtrade bot.
+    Please do not modify this strategy, it's  intended for internal use only.
+    Please look at the SampleStrategy in the user_data/strategy directory
+    or strategy repository https://github.com/freqtrade/freqtrade-strategies
+    for samples and inspiration.
+    """
+    INTERFACE_VERSION = 2
+
+    # Minimal ROI designed for the strategy
+    minimal_roi = {'0': 5}
+
+    # Optimal stoploss designed for the strategy
+    stoploss = -0.003
+
+    # Optimal ticker interval for the strategy
+    ticker_interval = '1h'
+
+    # Optional order type mapping
+    order_types = {
+        'buy': 'limit',
+        'sell': 'limit',
+        'stoploss': 'limit',
+        'stoploss_on_exchange': False
+    }
+
+    # Number of candles the strategy requires before producing valid signals
+    startup_candle_count: int = 30
+
+    # Optional time in force for orders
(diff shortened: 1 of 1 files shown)


user_data/strategies/berlinguyinca/TDSequentialStrategy.py
```diff
@@ -0,0 +1,132 @@
...
+    or strategy repository https://github.com/freqtrade/freqtrade-strategies
+    for samples and inspiration.
+    """
+    INTERFACE_VERSION = 2
+
+    # Minimal ROI designed for the strategy
+    minimal_roi = {'0': 5}
+
+    # Optimal stoploss designed for the strategy
+    stoploss = -0.003
```
xmatthias, line 22: puh - at first glance i must say, a stoploss of 0.3% seems very tight.

it's very likely to trigger in real markets (excluding highvolume pairs) almost instantly due to constant small market movements ...
will look good in backtesting though - where such small market movemens are not relevant ...

hroff-1902: I'm hyperopting it for roi and stoploss, to adjust to more reasonable values, to have it working out of the box with the default config.

xmatthias: well - default config - not necessarily - it's valid for a strategy to be applicable only to 1h timeframe ... so i'd leave that as is ... just the stoploss is ... concerning at best ;)

hroff-1902:
```
$ freqtrade hyperopt -c config_binance.json --timerange 20190301-20200301 --strategy TDSequentialStrategy --print-all -e 1000 --spaces stoploss roi trailing --hyperopt-loss SortinoHyperOptLossDaily --min-trades 20 --hyperopt SampleHyperOpt
...
Best result:

   739/1000:    476 trades. Avg profit   0.17%. Total profit  160.94478861 BTC (  80.39Σ%). Avg duration 2716.5 min. Objective: -0.51684

ROI table:
{0: 0.3402, 133: 0.0874, 685: 0.03447, 2123: 0}
Stoploss: -0.19143
Trailing stop:
{   'trailing_only_offset_is_reached': True,
    'trailing_stop': True,
    'trailing_stop_positive': 0.01023,
    'trailing_stop_positive_offset': 0.04953}
```

bmoulkaf: backtesting , and therefore hyperopting too , is really not reliable in here beacause if the low of the candle doesn't trigger the stoploss , then it takes for granted the stoploss from the high , which in this case is a huge difference seeing how tight it is .
also the sell strategy isn't effective since it is more of a short entry.
the strategy is just a direct application of the theory provided in the issue , and definitely needs a lot of tuning.

user_data/strategies/berlinguyinca/TDSequentialStrategy.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def populate_sell_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+        """
+        Based on TA indicators, populates the sell signal for the given dataframe
+        :param dataframe: DataFrame
+        :param metadata: Additional information, like the currently traded pair
+        :return: DataFrame with buy columnNA / NaN values
+        """
+        dataframe["sell"] = 0
+        dataframe.loc[((dataframe['exceed_high']) |
```
xmatthias, line 129: would you want & here too? (looking at the buy signal section - that would seem aligned (since the last commit, anyway) ?

bmoulkaf: well for the buy we have a signal and it's confirmation , but for the sell i thought it'd be better to exit as soon as we got one of them , but as i said it is more of a short entry , and it is extremely rare to have it right after a buy signal before trailing/stoploss gets triggered.
so it'd be better to tune the stoploss well , or find another sell strategy.

user_data/strategies/berlinguyinca/TDSequentialStrategy.py
hclivess: > puh - at first glance i must say, a stoploss of 0.3% seems very tight.
 ...
Actually 0.03%

bmoulkaf: pretty sure it is 0.3% ( 0.003 * 100 = 0.3%)

hroff-1902: 0.3% as stoploss is usually (for most of the markets) is very tight. Too tight, as anything below 1.0% -- it will cause strange behavior, especially for cheap shitcoins, where spread can be wider than this tight stoploss often...

hroff-1902: A hyperopting result for 1d timeframe:
```
Best result:

342/1000: 291 trades. Avg profit 2.40%. Total profit 1400.59209880 BTC ( 699.60Σ%). Avg duration 64.3 min. Objective: -38.52262

ROI table:
{0: 0.81149, 10074: 0.34659, 22300: 0.13231, 54710: 0}
Stoploss: -0.02046
Trailing stop:
{ 'trailing_only_offset_is_reached': False,
'trailing_stop': True,
'trailing_stop_positive': 0.01023,
'trailing_stop_positive_offset': 0.01168}
```

(Note avg duration -- it's only 64 min for 1d timeframe)
Stoploss is 0.02, which is not high, but is not as tight as 0.003.

user_data/strategies/berlinguyinca/TDSequentialStrategy.py
hroff-1902: Should not same ('&') be applied to the sell part?

user_data/strategies/berlinguyinca/TDSequentialStrategy.py
bmoulkaf: this still looks very promising !
i'm still trying to improve the strategy using the sources you shared on the issue , i think it can achieve even better results after this.

user_data/strategies/berlinguyinca/TDSequentialStrategy.py
xmatthias: @hroff-1902 see above ... there is a thought behind this :+1:

user_data/strategies/berlinguyinca/TDSequentialStrategy.py
hroff-1902: the results above -- for the variant with '|' between the conditions

PR #88, opened 2020-06-26
Hyperopts for optimization of indicators parameters

Nanidroid:
## Summary

1- An hyperopt for optimizing the Bollinger Bands parameters and RSI value of a simple strategy.
And one Hyperopt for optimizing the MACD parameters in a simple crossover signal strategy.

## Quick strategy idea

1- Buys when close price is under the lower bollinger band, sells when close price crosses one of the bands, uses RSI as guard.
2- Buys when MACD line crosses above the signal line and both are under 0. Sells when the MACD line crosses below the signal line and both lines are above 0.


user_data/hyperopts/BB_RSIhyperopt.py
```diff
@@ -0,0 +1,193 @@
...
+            # GUARDS AND TRENDS
+            if 'aboversi-enabled' in params and params['aboversi-enabled']:
+                conditions.append(dataframe['rsi'] > params['rsi-value'])
+            if 'belowrsi-enabled' in params and params['belowrsi-enabled']:
+                conditions.append(dataframe['rsi'] < params['rsi-value'])
+
+            # TRIGGERS
+            if 'trigger' in params:
+                if params['trigger'][2] == 'below':
+                    conditions.append(dataframe['close'] < dataframe[f"BB_lowerband({params['trigger'][0]},{params['trigger'][1]})"])
```
xmatthias: Is it intentional that "below" checks for close < bb_lower - while "above" checks for crossed_above() ? 

might be worth to add a 3rd variant, crossed-below to compare that too?

user_data/hyperopts/MACDCrossHyperopt.py
```diff
@@ -0,0 +1,182 @@
...
+            """
+            conditions = []
+            # TRIGGERS
+            if 'trigger' in params:
+                conditions.append(qtpylib.crossed_above(
+                    dataframe[f"MACD({params['trigger'][0]},{params['trigger'][1]},{params['trigger'][2]})"],
+                    dataframe[f"MACDsignal({params['trigger'][0]},{params['trigger'][1]},{params['trigger'][2]})"]) &
+
+                    (dataframe[f"MACD({params['trigger'][0]},{params['trigger'][1]},{params['trigger'][2]})"] < 0) &
+                    (dataframe[f"MACDsignal({params['trigger'][0]},{params['trigger'][1]},{params['trigger'][2]})"] < 0)
```
xmatthias, line 66: Wouldn't this "< 0" condition deserve another (additional) condition?

it seems like a guard mixed in ... a cross can also be valid > 0 ... i'd let hyperopt take this decision (maybe you tried and that does not work - looking for your input here!)

xmatthias:
As we now have support for hyperoptimization within the strategy via StrategyParameters, we won't need explicit hyperopt files any longer.

PR #96, opened 2020-08-25
correct typo

coorasse:


xmatthias (approved):
Great catch! Thanks for taking the time to correct this! 👍

PR #117, opened 2021-02-05
add hyperopts for BinHV45 strategy

JoeSchr:
## Summary
I thought I would share my hyperopt for BinHV45.

I actually have no idea what I'm doing, but since there is a need for it (ttps://github.com/freqtrade/freqtrade-strategies/issues/33) I thought I would just share what I got and the internet being the internet somebody would hopefully correct me.

Partly solve the issue: #33

### DIFF
--- a/user_data/hyperopts/BinHV45HyperOpt.py
+++ b/user_data/hyperopts/BinHV45HyperOpt.py
@@ -1,0 +1,104 @@
+# pragma pylint: disable=missing-docstring, invalid-name, pointless-string-statement
+
+# --- Do not remove these libs ---
+from functools import reduce
+from typing import Any, Callable, Dict, List
+
+import numpy as np  # noqa
+import pandas as pd  # noqa
+from pandas import DataFrame
+from skopt.space import Categorical, Dimension, Integer, Real  # noqa
+
+from freqtrade.optimize.hyperopt_interface import IHyperOpt
+
+# --------------------------------
+# Add your lib to import here
+import talib.abstract as ta  # noqa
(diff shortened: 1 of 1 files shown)


user_data/hyperopts/BinHV45HyperOpt.py
```diff
@@ -0,0 +1,104 @@
...
+    - The prototypes for the methods: populate_indicators, indicator_space, buy_strategy_generator.
+
+    The methods roi_space, generate_roi_table and stoploss_space are not required
+    and are provided by default.
+    However, you may override them if you need 'roi' and 'stoploss' spaces that
+    differ from the defaults offered by Freqtrade.
+    Sample implementation of these methods will be copied to `user_data/hyperopts` when
+    creating the user-data directory using `freqtrade create-userdir --userdir user_data`,
+    or is available online under the following URL:
+    https://github.com/freqtrade/freqtrade/blob/develop/freqtrade/templates/sample_hyperopt_advanced.py.
```
xmatthias, line 39: I'd change this to describe what the goal of this hyperopt is (and for which strategy ...).
While the name makes it clear to me, not everyone will know all strategies.

Also, this is just ONE way to optimize this strategy - others might also include disabling certain conditions completely - so maybe mention this as well.

JoeSchr: @xmatthias I improved it, hope that's what you had in mind. Otherwise feel free to edit, without going back and forth, if this is preferable for you

xmatthias (approved):
LGTM 👍 

Thanks a lot for sharing!

PR #122, opened 2021-02-12
Adding Stochastic + RSI + TEMA Strategy and Hyperopt

werkkrew:
## Summary
Just wanted to contribute.


werkkrew:
This is not a real strategy and was not written in the documented correct way.

xmatthias:
i think you misunderstood me in discord :)

It's not the documented way - that doesn't make it wrong per se - HOWEVER - if at the same time, backtest and hyperopt resuts don't align - then the "documented" way might be a first step - as it removes some complexity (even though it's more manual work).

PR #127, opened 2021-02-22
Hyperopt and Strategy "Swing-High-To-Sky"

OtenMoten:
## Hello dear community,

I like to share my newest hyperopt with you. I though about how cool it would be to know what's the perfect timeperiod for CCI indicator. In a strategy you do something like this: `dataframe['cci'] = ta.CCI(timeperiod=14)`

You would do this by hand for each timeperiod which is very annoying. Therefore, I created this hyperopt to looking for the perfect timeperiod for the CCI indicator. Please review this pull request very critical and share your minds.

Since the last two months (from 1st Jan 2021 until now) this strategy in BTC/USDT 30m chart had worked **very very** well. After two months I now optimize this strategy again.

## Summary
The goal of this hyper-optimization is to find the perfect timeframe of the CCI indicator (from 10 to 100) within a range from -400 to +400. The MACD indicator here is just a favorite of myself, replace with your favorit indicator if you like.

## IMPORTANT INFORMATION
Remember if you do trailing optimization you need to use cci names of the loops, i.e. "cci-40, cci-87, ...".

### DIFF
--- a/user_data/hyperopts/HO-SwingHighToSky.py
+++ b/user_data/hyperopts/HO-SwingHighToSky.py
@@ -1,0 +1,124 @@
+# pragma pylint: disable=missing-docstring, invalid-name, pointless-string-statement
+
+import talib.abstract as ta
+from pandas import DataFrame
+from typing import Dict, Any, Callable, List
+from functools import reduce
+
+import numpy as np
+from skopt.space import Categorical, Dimension, Integer, Real
+import freqtrade.vendor.qtpylib.indicators as qtpylib
+from freqtrade.optimize.hyperopt_interface import IHyperOpt
+
+__author__      = "Kevin Ossenbrück"
+__copyright__   = "Free For Use"
+__credits__     = ["Bloom Trading, Mohsen Hassan"]
+__license__     = "MIT"
+__version__     = "1.0"
+__maintainer__  = "Kevin Ossenbrück"
+__email__       = "kevin.ossenbrueck@pm.de"
+__status__      = "Live"
+
+cciTimeMin = 10
+cciTimeMax = 100
+cciValueMin = -400
+cciValueMax = 400
+cciTimeRange = range(cciTimeMin, cciTimeMax)
+
+class_name = 'HOSwingHighToSky'
+class HOSwingHighToSky(IHyperOpt):
+
+    @staticmethod
+    def populate_indicators(dataframe: DataFrame, metadata: dict) -> DataFrame:
+
+        macd = ta.MACD(dataframe)
+        dataframe['macd'] = macd['macd']
+        dataframe['macdsignal'] = macd['macdsignal']
+        
+        for cciTime in cciTimeRange:
+        
+            cciName = "cci-" + str(cciTime) 
+            dataframe[cciName] = ta.CCI(dataframe, timeperiod = cciTime)
+            
+        return dataframe
+
+    @staticmethod
+    def buy_strategy_generator(params: Dict[str, Any]) -> Callable:
+   
+        def populate_buy_trend(dataframe: DataFrame, metadata: dict) -> DataFrame:
+        
+            conditions = []
+                
+            # TRIGGERS & GUARDS
+            if 'trigger' in params:
+            
+                for cciTime in cciTimeRange:
+                
+                    cciName = "cci-" + str(cciTime)
+            
+                    if params['trigger'] == cciName:
+                        conditions.append(dataframe[cciName] < params["buy-cci-value"])
+                        conditions.append(dataframe['macd'] > dataframe['macdsignal'])
+                        conditions.append(dataframe['volume'] > 0)
+                
+            if conditions:
+                dataframe.loc[reduce(lambda x, y: x & y, conditions), 'buy'] = 1
+
+            return dataframe
+
+        return populate_buy_trend
+
+    @staticmethod
+    def indicator_space() -> List[Dimension]:
+        
+        buyTriggerList = []
+        
+        for cciTime in cciTimeRange:
+        
+            cciName = "cci-" + str(cciTime)
+            buyTriggerList.append(cciName)
+        
+        return [
+            Integer(cciValueMin, cciValueMax, name='buy-cci-value'),
+            Categorical(buyTriggerList, name='trigger')
+        ]
+
+    @staticmethod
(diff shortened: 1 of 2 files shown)


user_data/strategies/Swing-High-To-Sky.py
```diff
@@ -0,0 +1,71 @@
...
+import numpy # noqa
+
+__author__      = "Kevin Ossenbrück"
+__copyright__   = "Free For Use"
+__credits__     = ["Bloom Trading, Mohsen Hassan"]
+__license__     = "MIT"
+__version__     = "1.0"
+__maintainer__  = "Kevin Ossenbrück"
+__email__       = "kevin.ossenbrueck@pm.de"
+__status__      = "Live"
```
xmatthias, line 19: While i love that you specify author and everything, i'd want to move these into the docstring of the class.

That way, they won't show up as Syntax errors / linting errors in editors people use - and it's really tied to the strategy, not to the file.

The problem with moving this into the file's "top elvel" is that it'll be executed every time the file is loaded, which (in the worst case) can have odd side-effects.

user_data/strategies/Swing-High-To-Sky.py
```diff
@@ -0,0 +1,71 @@
...
+
+class_name = 'SwingHighToSky'
```
xmatthias, line 21: that's not really needed -  and the class-name is right below anyway - so maybe i miss the point of this variable?

user_data/hyperopts/HO-SwingHighToSky.py
```diff
@@ -0,0 +1,124 @@
...
+from freqtrade.optimize.hyperopt_interface import IHyperOpt
+
+__author__      = "Kevin Ossenbrück"
+__copyright__   = "Free For Use"
+__credits__     = ["Bloom Trading, Mohsen Hassan"]
+__license__     = "MIT"
+__version__     = "1.0"
+__maintainer__  = "Kevin Ossenbrück"
+__email__       = "kevin.ossenbrueck@pm.de"
+__status__      = "Live"
```
xmatthias, line 20: Same as below on the strategy

user_data/hyperopts/HO-SwingHighToSky.py
```diff
@@ -0,0 +1,124 @@
...
+            if 'sell-trigger' in params:
+            
+                for cciTime in cciTimeRange:
+                
+                    cciName = "cci-" + str(cciTime)
+            
+                    if params['sell-trigger'] == cciName:
+                        conditions.append(dataframe[cciName] > params["sell-cci-value"])
+                        conditions.append(dataframe['macd'] < dataframe['macdsignal'])
+
```
xmatthias, line 103: I think you will want to check "volume > 0" here too unless there's a good reason why you don't do that intentionally.
(best check the sample_hyperopt on how to do that in hyperopt).


Ideally, results will not change - however i've seen this as problematic on low-volume pairs which didn't have trades for HOURS ... backtesting / hyperopt then simulates buys / sells at these candles, which however have almost 0 chance of ever actually filling when using dry-run / live.

Obviously the same applies to populate_buy (for both strategy and hyperopt - in both instances).

xmatthias (commented):
Overall - looks great! :+1: 

Just a few minor things / potential improvements ... 
and one request - can we name the files the same way the class is called (so `Swing-High-To-Sky.py` becomes `SwingHighToSky.py`. 

That makes identifying the strategy easier than when the name mismaches (i know that's not followed consistently at the moment - but i'm trying to work on that ....)

user_data/strategies/Swing-High-To-Sky.py
```diff
@@ -0,0 +1,71 @@
...
+        return []
+
+    def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+        
+        macd = ta.MACD(dataframe)
+        dataframe['macd'] = macd['macd']
+        dataframe['macdsignal'] = macd['macdsignal']
+        dataframe['macdhist'] = macd['macdhist']
+        
+        dataframe['cci'] = ta.CCI(dataframe)
```
xmatthias, line 47: i find the strategy part here quite missleading - optimizing buy and sell space i get 2 different cci timeframes (that's fine) - however the strategy does not indicate this directly - making alignment of the strategy quite difficult.


```suggestion
        dataframe['cci-buy'] = ta.CCI(dataframe, timeperiod=xx)
        dataframe['cci-sell'] = ta.CCI(dataframe, timeperiod=xx-sell)
```

user_data/strategies/Swing-High-To-Sky.py
```diff
@@ -0,0 +1,71 @@
...
+
+        return dataframe
+
+    def populate_buy_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+        
+        dataframe.loc[
+            (
+                (dataframe['macd'] > dataframe['macdsignal']) &
+                (dataframe['cci'] <= -100.0)
```
xmatthias, line 56: to fully align, you should add `dataframe['volume'] > 0` here too - it's part of hyperopt - and can slightly change results depending on the pairs (and exchange) used

user_data/strategies/Swing-High-To-Sky.py
OtenMoten: Could be removed. Just typed in my mail if somebody has urgents reasons to contact me.

user_data/hyperopts/HO-SwingHighToSky.py
OtenMoten: Same as above.

user_data/hyperopts/HO-SwingHighToSky.py
OtenMoten: I thought that volume information is not included in historical data - isn't it? Therefore, it's excluded in the hyperopt but included in strategy. But as I can see here `conditions.append(dataframe['volume'] > 0)` everything is fine and I will know in future.

Thanks for the note.

user_data/strategies/Swing-High-To-Sky.py
OtenMoten: Jep, perfect here.

OtenMoten (commented):
Approved.

xmatthias (approved):
LGTM :+1: 

Thanks for taking the time to contribute this!

PR #128, opened 2021-02-23
Corrected ADXMomentum.py

XL-Reaper:
There was still one more error in relation to the mentioned original strategy. I fixed it to check the correct variable.


xmatthias (approved):
Nice catch 👍

PR #130, opened 2021-03-03
Add TrailingSL as an example for using ATR or any other indicator inside of custom_stoploss

JoeSchr:
I created this for https://github.com/freqtrade/freqtrade/pull/4478, so I could test requested changes with `freqtrade backtesting` and `freqtrade trade --dry-run` with it.

I figured I could share it as an full example here as well.
This PR should wait until https://github.com/freqtrade/freqtrade/pull/4478 is resolved.

ROADBLOCKS before undraft:
- [x] ERROR: `freqtrade.exceptions.StrategyError: datetime.datetime(2021, 3, 4, 14, 6, 28, 715919, tzinfo=datetime.timezone.utc)`
- [x] NameError: name 'datetime' is not defined


JoeSchr:
wait until https://github.com/freqtrade/freqtrade/pull/4478 is worekd out.
todo:
- [ ] NameError: name 'datetime' is not defined
.

user_data/strategies/trailing_sl.py
```diff
@@ -0,0 +1,411 @@
...
+                # backtesting/hyperopt
+                if self.dp.runmode.value in ('backtest', 'hyperopt'):
+                    relative_sl = self.custom_info[pair].loc[current_time][SL_INDICATOR_NAME]
+                # for live, dry-run, storing the dataframe is not really necessary, 
+                # it's available from get_analyzed_dataframe()
+                else:
+                    # so we need to get analyzed_dataframe from dp
+                    dataframe, last_updated = self.dp.get_analyzed_dataframe(pair=pair,
+                                                                             timeframe=self.timeframe)
+                    relative_sl = dataframe[last_updated][SL_INDICATOR_NAME]
```
xmatthias, line 49: this line will be the breaking factor.

The dataframe is 
a) not indexed by date, but by numbers
b) last_updated will NOT include the candle date, but the time of the last analysis - which is in every case at least a few microseconds after the candle.

Best use `.last()` or so to get the latest candle (but with a comment saying that this is ONLY done because this is not used by backtesting).

JoeSchr: Ok, if this is not an `DatetimeIndex`, `dataframe.iloc[-1]` should do the same job then `.last(-1)` right?

xmatthias: yes, .iloc[-1] will do the same thing

JoeSchr: I fixed this. 
- I didn't use `last()` because it's only available for `DatetimeIndex` 
- I didn't use `.iloc[-1]` because there are a lot of stern warnings in the docs do don't use it and I don't want to confuse less technically aligned user by now using it in an "official" sample
- I did use `.iat[-1]` because it's supposed to be fasted by working on the Series of the requested column directly

user_data/strategies/trailing_sl.py
```diff
@@ -0,0 +1,414 @@
...
+import talib.abstract as ta
+import freqtrade.vendor.qtpylib.indicators as qtpylib
+from datetime import datetime
+from freqtrade.persistence import Trade
+from freqtrade.state import RunMode
+
+class TrailingSL(IStrategy):
+    """
+    this is an abstract stump class, just implmenting `custom_stoploss`
+    you are supposed to inherit from it in your own strategy, e.g.:
```
xmatthias, line 21: i'm not sure if doing it this way (via an intermediat class users should inherit from) is doing people a great favor.

It'll make this seem overly complex to inexperienced users, while in reality it's just the custom stoploss + a few lines of extra in the populate_indicator methods.

JoeSchr: Ha, I was working at that while you wrote this. 

I only wrote it this way because I couldn't importing an inherited class from a another file working and didn't loose any more time. 
I tried by using
`from .trailing_sl import TrailingSL`
but it couldn't be resolved.

now I gave that another goal, because I thought, cwd is probably root of projects, so I tried
`from user_data/strategies/trailing_sl import TrailingSL` and it also isn't resolved. How can I import one strategy into another. Docs only say how to use them from the same file.

I understand your request is, to simplify it even more and just give `TrailingSL` a really simple `buy` method, but I would still like to know how to do above, if you got a minute.

Also I added a line, so `custom_info` is only populated in backtesting and hyperopt. that should keep memory down on live/dry-run

xmatthias: https://www.freqtrade.io/en/latest/strategy-advanced/#derived-strategies ? (i know noone reads the info boxes, but ... ) :laughing:

JoeSchr:
You are completely right. You can't imagine how perplexed I'm right now.

I must have looked at that page at least 2-3 times the last few days. And I
pride myself of being a careful reader (which some of my more anal PRs
prove). Not only did I look at it via the docs website, also via the raw
.MD files in my editor.

This box absolutely didn't register in my brain. I could have sworn it
wasn't there. Amazing.

Thank you


__

Typed on a mobile, please excuse any typing mistakes or short-spokeness.

user_data/strategies/inherit_trailing_stoploss.py
```diff
@@ -0,0 +1,363 @@
...
+# Add your lib to import here
+import talib.abstract as ta
+import freqtrade.vendor.qtpylib.indicators as qtpylib
+from datetime import datetime
+from freqtrade.persistence import Trade
+from freqtrade.state import RunMode
+from user_data.strategies.trailing_sl import TrailingSL
+
+# This class is a sample. Feel free to customize it.
+class InheritTrailingStoploss(IStrategy):
```
xmatthias, line 20: if you import it - you should use the Super class as well ...

JoeSchr: > if you import it - you should use the Super class as well 
you mean: `class InheritTrailingStoploss(TrailingSL):`

> in general though - i still think both classes should be combined into one.
Yeah I will simplify it into one class for the `custom_stoploss` example. give it some sort of placeholder buy signal so it could be run. then maybe still keep the `user_data/strategies/inherit_trailing_stoploss.py` file around, as an additional example for how to import from a different file, since other people might as well be incapable of reading info boxes :disappointed:

JoeSchr:
I simplified the example as requested and remove the inherited example for now

user_data/strategies/custom_stoploss_with_psar.py
```diff
@@ -0,0 +1,84 @@
...
+    def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+        dataframe['sar'] = ta.SAR(dataframe)
+        if self.dp.runmode.value in ('backtest', 'hyperopt'):
+            self.custom_info[metadata['pair']] = dataframe[['date', 'sar']].copy().set_index('date')
+
+        # all "normal" indicators:
+        dataframe['rsi'] = ta.RSI(dataframe)
+        return dataframe
+
+    def populate_buy_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
```
xmatthias, line 71: to have a (working) strategy, you'll also need to implement the populate_sell_trend method ...

JoeSchr: added `populate_sell_trend` :facepalm: 

also remove RSI indicator, just using SAR to make the example simpler

xmatthias (approved):
LGTM :+1: 

i hope people can put this to good use and benefit from it :+1:

JoeSchr:
> LGTM 
> 
> i hope people can put this to good use and benefit from it

Me too :hugs:

JoeSchr:
TLDR: is a qualifier needed to access `custom_info dataframe` at or near `trade.open_date`, because open candle might still be open when `custom_stoploss()` is called?

I have to revisit this PR with a question, regarding the use of DatetimeIndex while in `custom_stoploss()`.

I would like to add to this an example on how to access information for `trade.open_date`, so there is a best practice for reference. I would like to avoid to make it unneeded convoluted, because I think it's more confusing for beginnners. 

But I question, what happens if somebody would like to use `trade.open_date` to access `custom_info` while we are still processing the ticks while inside the buy signal candle. especially in dry/live mode. 
depending on the timeframe (e.g. 1h), this could be quite some issues, if somebody tries to do something like this:

```python
   custom_info...loc[trade.open_date]['sar']
```

maybe it would need to be more like this.
```python
custom_info.loc[(df['date'] >= trade.open_date) & (df['date'] < trade.open_date+self.config.timeframe)]['sar']
```

what are your thoughts @xmatthias?

JoeSchr:
Thanks!

I think I can still build something useful within this constraints. 

I'm going to share it in a different PR then though, except if you want me to reuse this one?

xmatthias:
it's merged, you can't reuse this PR, it's merged.

PR #132, opened 2021-03-10
fix hyperopting multi-value categories on AverageHyperopt

FelipeFernandesLeandro:
fix this [issue](https://github.com/freqtrade/freqtrade/issues/3796#issue-706185121)
with [xiedidan](https://github.com/xiedidan) [solution](https://github.com/freqtrade/freqtrade/issues/3796#issuecomment-749971896)

### DIFF
--- a/user_data/hyperopts/AverageHyperopt.py
+++ b/user_data/hyperopts/AverageHyperopt.py
@@ -43,9 +43,11 @@ class AverageHyperopt(IHyperOpt):
             conditions = []
             # TRIGGERS
             if 'trigger' in params:
+                trigger = [int(item) for item in params['trigger'].split('-')]
+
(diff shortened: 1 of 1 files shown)


user_data/hyperopts/AverageHyperopt.py
```diff
@@ -114,7 +118,7 @@ def sell_indicator_space() -> List[Dimension]:
             for medium in range(mediumRangeBegin, mediumRangeEnd):
                 # The output will be (short, long)
```
xmatthias, line 119: I think the comment should also be changed to explain why it's used as string ... 

otherwise it'll be rather unclear for people reading this in a few weeks / months ...

xmatthias (approved):
LGTM :+1: 

Thanks!

PR #134, opened 2021-03-15
Create HO-Strategy005.py and update Swing-High-To-Sky hyperopt and strategy.

OtenMoten:
## Summary
Regarding to `Strategy005.py` in https://github.com/freqtrade/freqtrade-strategies/blob/master/user_data/strategies/Strategy005.py I've created a hyperopt for this strategy. 

## Upcoming
I will manage this hyperopt actively because I run it live.

## Solving an issue
In the strategy from https://github.com/mishaker, special thanks to @mishaker, there are certain parameter values inserted and I like to get the most perfect values based on the pairs in my config file in a certain timerange.

## Special notes
The strategy from @mishaker and my hyperopt is designed for the 5m timeframe.

### DIFF
--- a/user_data/hyperopts/HO-Strategy005.py
+++ b/user_data/hyperopts/HO-Strategy005.py
@@ -45,20 +45,17 @@ minusdiValueMax = 100
 fishRsiNormaValueMin = 1
 fishRsiNormaValueMax = 100
 
-class HODobby(IHyperOpt):
 
+class HODobby(IHyperOpt):
     """
-    If you trade on Binance then the API endopoint is "api.binance.com".
-    It's based in Tokyo. You can get a VPS in Tokyo on Vultr with 2ms latency.
-    I feel free to share my referral link (you get a bonus too):
-        > https://www.vultr.com/?ref=8806640
+    Hyperopt file for Strategy005
     """
-    
+
     ############### THIS STRATEGY IS DESIGNED FOR 5m TIMEFRAME ###############
-    
+
     @staticmethod
     def populate_indicators(dataframe: DataFrame, metadata: dict) -> DataFrame:
-        
+
         # MACD
         # tadoc.org/indicator/MACD.htm
         macd = ta.MACD(dataframe)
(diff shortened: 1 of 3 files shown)


user_data/hyperopts/HO-Strategy005.py
```diff
@@ -0,0 +1,171 @@
...
+fishRsiNormaValueMin = 1
+fishRsiNormaValueMax = 100
+
+class HODobby(IHyperOpt):
+
+    """
+    If you trade on Binance then the API endopoint is "api.binance.com".
+    It's based in Tokyo. You can get a VPS in Tokyo on Vultr with 2ms latency.
+    I feel free to share my referral link (you get a bonus too):
+        > https://www.vultr.com/?ref=8806640
```
xmatthias, line 54: i would like to not put referal links into strategies / hyperopt files.

we don't have them anywhere else (as far as i know) - and by incorporating these in the repository, it would mean that we recommend them - which i can't (because i never used them).

user_data/hyperopts/HO-Strategy005.py
```diff
@@ -0,0 +1,171 @@
...
+                dataframe.loc[reduce(lambda x, y: x & y, conditions), 'buy'] = 1
+
+            return dataframe
+
+        return populate_buy_trend
+
+    @staticmethod
+    def indicator_space() -> List[Dimension]:
+        
+        buyTriggerList = ["True"]
```
xmatthias, line 122: Don't do this ... but instead remove / comment this from the space if it only contains one value 

While you'll think it's just one option, so it'll not do anything - scikit-optimize will still consider it as a value to optimize - and it will therefore consume CPU.

xmatthias (changes requested):
What i'm kindof missing is the actual strategy this goes along with ...

is it still with Stragtegy005, or was that just an inspiration and it's really a derivative (which will also require a different strategy file?)

If it's for Strategy005, maybe put that in the docstring - just so it's clear (I've been reading the comments over and over, and it really wasn't clear.

user_data/hyperopts/HO-Strategy005.py
OtenMoten: You are absolute right.

OtenMoten: But's it's pretty cool hoster :-)

user_data/hyperopts/HO-Strategy005.py
OtenMoten: OK, let's remove it.

OtenMoten:
Updated Swing-High-To-Sky hyperopt and strategy file.

OtenMoten:
> What i'm kindof missing is the actual strategy this goes along with ...
> 
> is it still with Stragtegy005, or was that just an inspiration and it's really a derivative (which will also require a different strategy file?)
> 
> If it's for Strategy005, maybe put that in the docstring - just so it's clear (I've been reading the comments over and over, and it really wasn't clear.

I already did this on line 24 to 26.

xmatthias (approved):
i did remove the referral code ... simply marking the conversation to "resolved" does not remove it from the file ... 

LGTM :+1:

PR #135, opened 2021-03-15
Add an example how to use custom_info[trade.open_date] in custom_stoploss()

JoeSchr:
by implementing a fixed risk/reward ratio.

## Quick strategy idea

    This strategy uses custom_stoploss() to enforce a fixed risk/reward ratio
    by first calculating a dynamic initial stoploss via ATR - last negative peak

    After that, we caculate that initial risk and multiply it with an risk_reward_ratio
    Once this is reached, stoploss is set to it and sell signal is enabled

    Also there is a break even ratio. Once this is reached, the stoploss is adjusted to minimize
    losses by setting it to the buy rate + fees.

### DIFF
--- a/user_data/strategies/fixed_riskreward_loss.py
+++ b/user_data/strategies/fixed_riskreward_loss.py
@@ -45,7 +45,7 @@ class FixedRiskRewardLoss(IStrategy):
             custom_stoploss using a risk/reward ratio
         """
         result = break_even_sl = takeprofit_sl = -1
-        custom_info_pair = self.custom_info[pair]
+        custom_info_pair = self.custom_info.get(pair)
         if custom_info_pair is not None:
             # using current_time/open_date directly via custom_info_pair[trade.open_daten]
             # would only work in backtesting/hyperopt.
--- a/user_data/strategies/fixed_riskreward_loss.py
+++ b/user_data/strategies/fixed_riskreward_loss.py
@@ -87,9 +87,8 @@ class FixedRiskRewardLoss(IStrategy):
         return result
 
     def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
-        dataframe['min'] = dataframe['low'].rolling(48).min()
         dataframe['atr'] = ta.ATR(dataframe)
-        dataframe['stoploss_rate'] = dataframe['min']-(dataframe['atr'])
+        dataframe['stoploss_rate'] = dataframe['close']-(dataframe['atr']*2)
         self.custom_info[metadata['pair']] = dataframe[['date', 'stoploss_rate']].copy().set_index('date')
 
         # all "normal" indicators:
@@ -116,6 +115,6 @@ class FixedRiskRewardLoss(IStrategy):
         :return: DataFrame with buy column
         """
 
-        # Always sells
-        dataframe.loc[:, 'sell'] = 1
+        # Never sells
+        dataframe.loc[:, 'sell'] = 0
         return dataframe
--- a/user_data/strategies/fixed_riskreward_loss.py
+++ b/user_data/strategies/fixed_riskreward_loss.py
@@ -35,8 +35,6 @@ class FixedRiskRewardLoss(IStrategy):
     }
     use_custom_stoploss = True
     stoploss = -0.9
-    sell_profit_only = True
-    sell_profit_offset = 1 # 100%, get's set dynamically in trail
 
     def custom_stoploss(self, pair: str, trade: 'Trade', current_time: datetime,
                         current_rate: float, current_profit: float, **kwargs) -> float:
@@ -55,7 +53,6 @@ class FixedRiskRewardLoss(IStrategy):
 
             # trade might be open too long for us to find opening candle
             if(len(open_df) != 1):
-                self.sell_profit = False # re-activate sell signal at any profit
                 return -1 # won't update current stoploss
 
             initial_sl_abs = open_df['stoploss_rate']
@@ -86,9 +83,6 @@ class FixedRiskRewardLoss(IStrategy):
             if(current_profit >= take_profit_pct):
                 takeprofit_sl = take_profit_price_abs/current_rate-1
                 result = takeprofit_sl
-
-            # enable sell signal only after take_profit treshold is reached
-            self.sell_profit_offset = take_profit_pct
 
         return result
 
--- a/user_data/strategies/fixed_riskreward_loss.py
+++ b/user_data/strategies/fixed_riskreward_loss.py
@@ -1,0 +1,123 @@
+# pragma pylint: disable=missing-docstring, invalid-name, pointless-string-statement
+# isort: skip_file
+# --- Do not remove these libs ---
+import numpy as np  # noqa
+import pandas as pd  # noqa
+from pandas import DataFrame
+
+from freqtrade.strategy.interface import IStrategy
+
+# --------------------------------
+# Add your lib to import here
+import talib.abstract as ta
+import freqtrade.vendor.qtpylib.indicators as qtpylib
+from datetime import datetime
+from freqtrade.persistence import Trade
+from freqtrade.state import RunMode
+import logging
+logger = logging.getLogger(__name__)
+
+class FixedRiskRewardLoss(IStrategy):
+    """
+    This strategy uses custom_stoploss() to enforce a fixed risk/reward ratio
+    by first calculating a dynamic initial stoploss via ATR - last negative peak
+
+    After that, we caculate that initial risk and multiply it with an risk_reward_ratio
+    Once this is reached, stoploss is set to it and sell signal is enabled
+
+    Also there is a break even ratio. Once this is reached, the stoploss is adjusted to minimize
+    losses by setting it to the buy rate + fees.
+    """
+
+    custom_info = {
+        'risk_reward_ratio': 3.5,
+        'set_to_break_even_at_profit': 1,
+    }
+    use_custom_stoploss = True
+    stoploss = -0.9
+    sell_profit_only = True
+    sell_profit_offset = 1 # 100%, get's set dynamically in trail
+
+    def custom_stoploss(self, pair: str, trade: 'Trade', current_time: datetime,
+                        current_rate: float, current_profit: float, **kwargs) -> float:
+
+        """
+            custom_stoploss using a risk/reward ratio
+        """
+        result = break_even_sl = takeprofit_sl = -1
+        custom_info_pair = self.custom_info[pair]
+        if custom_info_pair is not None:
+            # using current_time/open_date directly via custom_info_pair[trade.open_daten]
+            # would only work in backtesting/hyperopt.
+            # in live/dry-run, we have to search for nearest row before it
+            timezone = custom_info_pair.index.tz
+            open_date = trade.open_date.replace(tzinfo=timezone)
+            open_date_mask = custom_info_pair.index.unique().get_loc(open_date, method='ffill')
+            open_df = custom_info_pair.iloc[open_date_mask]
+            initial_sl_abs = open_df['stoploss_rate']
+
+            # calculate initial stoploss at open_date
+            initial_sl = initial_sl_abs/current_rate-1
+
+            # calculate take profit treshold
+            # by using the initial risk and multiplying it
+            risk_distance = trade.open_rate-initial_sl_abs
+            reward_distance = risk_distance*self.custom_info['risk_reward_ratio']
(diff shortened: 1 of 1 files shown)


user_data/strategies/fixed_riskreward_loss.py
```diff
@@ -0,0 +1,123 @@
...
+            custom_stoploss using a risk/reward ratio
+        """
+        result = break_even_sl = takeprofit_sl = -1
+        custom_info_pair = self.custom_info[pair]
+        if custom_info_pair is not None:
+            # using current_time/open_date directly via custom_info_pair[trade.open_daten]
+            # would only work in backtesting/hyperopt.
+            # in live/dry-run, we have to search for nearest row before it
+            timezone = custom_info_pair.index.tz
+            open_date = trade.open_date.replace(tzinfo=timezone)
```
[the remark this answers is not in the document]
JoeSchr: Thanks for your inputs!

Will fix the other ones, but should I just assume people are using a recent develop? It would be less verbose, I guess. 

How soon to you plan to merge current develop into stable?

user_data/strategies/fixed_riskreward_loss.py
```diff
@@ -0,0 +1,123 @@
...
+            open_date_mask = custom_info_pair.index.unique().get_loc(open_date, method='ffill')
+            open_df = custom_info_pair.iloc[open_date_mask]
+            initial_sl_abs = open_df['stoploss_rate']
```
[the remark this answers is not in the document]
JoeSchr: Totally.

That's the reason a "real" initial_stoploss when combined with `custom_stoploss` would be great, because then this info would just persist as long as the trade is active.

I just gave up yesterday bashing my head against this, but I will add a simple guard today which just `return 1` & resetting profit iffset then, assuming if the market went side for a long time, initially buy signal was off...

user_data/strategies/fixed_riskreward_loss.py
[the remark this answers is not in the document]
xmatthias: usually, i try to release a new version towards the end of the month - unless it's something really urgent (needs fixing) in stable - or something is broken in develop (can delay a release).

user_data/strategies/fixed_riskreward_loss.py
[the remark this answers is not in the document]
xmatthias: well wait, that's a different usecase.
your custom stoploss will be called with available data for the first 500(ish) candles after the trade is open - so the only scenario the initial stoploss would not be overwritten by custom stoploss is if you opened the trade, and IMMEDIATELY (before one iteration completes) stop the bot for 500+ candles, only then to relaunch this.

So in reality - the stop will update correctly - it'll just "stop" to update after the candles run out. It's not a common scenario for most strategies (usually, if you aim for multi-day trades, you'll not use 5m candles) i'd hope ... just one thing to keep in mind.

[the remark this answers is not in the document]
JoeSchr: Maybe my understanding how and when `trade.initial_stoploss` gets set it flawed. but my understanding is, that it will never be set by `custom_stoploss` just by `self.stoploss` (from config/strategy) or `force_stoploss` (in hyperopt). 

that's what I mean with "real initial_stoploss". because when using `custom_stoploss` like in this example, if you look into `trade.initial_stoploss` you will always just see the placeholder value (setting `self.stoploss` to be wide enough so it doesn't interfere with `custom_stoploss`).

so for calculating a fixed risk/reward ratio inside `custom_stoploss()`, `trade.initial_stoploss` is therefor not particulatly usefull, because it will always be `-0.99` (or whatever is used as placeholder value)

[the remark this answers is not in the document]
JoeSchr: > so the only scenario the initial stoploss would not be overwritten by custom stoploss i
ah, now I get it. you didn't mean that initial stoploss would get updated by custom_stoploss per set (`initial_stoploss = custom_stoploss()`) but that the stoploss set (enforced by exchange or bot) would be different off `initial_stoploss` once `custom_stoploss()` would be called at least once. 

Sorry, I let my hope flare up, that it would actually get initialized by `custom_stoploss()` and I just missed it so far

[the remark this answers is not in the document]
JoeSchr: I just don't get what usecase you are refering to. If I `return 1` from `custom_stoploss()` it just would keep current stoploss. 

Which I find a good behavior once we are in uncharted territory, ain't it?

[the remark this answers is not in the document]
xmatthias: i think you're overthinking it.

it's quite simple - if `initial_stoploss != stoploss` - then we're in "trailing" territory (the stop was moved at least once).
custom-stoploss will not be able to set initial-stoploss - but you shoulnd't set it to 99% either, as then your full stake is at risk.
I assume the custom stop logic mainly comes into play once we're in "positive" territory - at which point the static stop value is irrelevant anyway.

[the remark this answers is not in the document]
JoeSchr: > i think you're overthinking it.

that's something I tend to do :D 

> it's quite simple - if `initial_stoploss != stoploss` - 

understood now

>then we're in "trailing" territory (the stop was moved at least once).

what significance does "trailing" territory have for us?

> custom-stoploss will not be able to set initial-stoploss - but you shoulnd't set it to 99% either, as then your full stake is at risk.

I kind of need to widen it enough to be able to use a dynamic stoploss inside custom_stoploss() though

> I assume the custom stop logic mainly comes into play once we're in "positive" territory - at which point the static stop value is irrelevant anyway.

ah no, because it's using `custom_info[open_date]['stoploss_rate']` to set a dynamic stoploss in the first place (using something like e.g. 2 times ATR)
```py
initial_sl_abs = open_df['stoploss_rate'] # line#57
...
initial_sl = initial_sl_abs/current_rate-1
...
result = initial_sl # line#77
```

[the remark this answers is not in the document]
JoeSchr: I added your requested changes. I think they are alright for now?  

we might want to table the discussion about `custom_stoploss() <-> initial_stoploss` until I come around to making a separate PR for it. It might be easier to discuss what I want to do when I provide specific code changes...

JoeSchr:
Should I also simplify this further and add as an example to `custom_stoploss` documentation?

user_data/strategies/fixed_riskreward_loss.py
```diff
@@ -0,0 +1,127 @@
...
+
+            # break_even tries to set sl at open_rate+fees (0 loss)
+            break_even_profit_distance = risk_distance*self.custom_info['set_to_break_even_at_profit']
+            # break_even gets triggerd at this profit
+            break_even_profit_pct = (break_even_profit_distance+current_rate)/current_rate-1
+
+            result = initial_sl
+            if(current_profit >= break_even_profit_pct):
+                break_even_sl = (trade.open_rate*(1+trade.fee_open+trade.fee_close) / current_rate)-1
+                result = break_even_sl
```
xmatthias, line 84: maybe i'm wrong ... but assuming my "break even profit" is 1% - and current profit is 1% - won't this (try) to set the stoploss at current price?

user_data/strategies/fixed_riskreward_loss.py
```diff
@@ -0,0 +1,127 @@
...
+class FixedRiskRewardLoss(IStrategy):
+    """
+    This strategy uses custom_stoploss() to enforce a fixed risk/reward ratio
+    by first calculating a dynamic initial stoploss via ATR - last negative peak
+
+    After that, we caculate that initial risk and multiply it with an risk_reward_ratio
+    Once this is reached, stoploss is set to it and sell signal is enabled
+
+    Also there is a break even ratio. Once this is reached, the stoploss is adjusted to minimize
+    losses by setting it to the buy rate + fees.
```
xmatthias, line 29: Do i get this right - once break-even-ratio is reached, you'll set stoploss to "break even" ? 

i think you might want to have an offset too - if i want "break even" to be reached after 5% - i would want the stop to be set to 1% absolute profit - not 0% profit ... (i know, that's strictly not break even anymore)

user_data/strategies/fixed_riskreward_loss.py
```diff
@@ -0,0 +1,127 @@
...
+    """
+
+    custom_info = {
+        'risk_reward_ratio': 3.5,
+        'set_to_break_even_at_profit': 1,
```
xmatthias, line 34: this sample value ... once i reach 100% profit, i set the stoploss to 0% profit? (so at max, i'd lose the 100% profit ...?)

user_data/strategies/fixed_riskreward_loss.py
```diff
@@ -0,0 +1,127 @@
...
+            if(current_profit >= break_even_profit_pct):
+                break_even_sl = (trade.open_rate*(1+trade.fee_open+trade.fee_close) / current_rate)-1
+                result = break_even_sl
+
+            if(current_profit >= take_profit_pct):
+                takeprofit_sl = take_profit_price_abs/current_rate-1
+                result = takeprofit_sl
+
+            # enable sell signal only after take_profit treshold is reached
+            self.sell_profit_offset = take_profit_pct
```
xmatthias, line 91: this will not work correctly at the moment.

This value is transferred to the configuration, and not used after loading the strategy - so this is a line without any effect.

user_data/strategies/fixed_riskreward_loss.py
JoeSchr: > maybe i'm wrong ... but assuming my "break even profit" is 1% - and current profit is 1% - won't this (try) to set the stoploss at current price?

correct

user_data/strategies/fixed_riskreward_loss.py
JoeSchr: correct

user_data/strategies/fixed_riskreward_loss.py
JoeSchr: oh, that's a pitty :(

would it be a major breaking change to adjust it so it's loaded after as well?

JoeSchr:
thanks @xmatthias for reviewing this PR. I will try to clarify. 
(Maybe this video does it better by visualizing it: https://youtu.be/Hgvt3lP7WxQ?t=140 (I timestamped it))

what this is for, is using a fixed risk/reward ratio while setting a dynamically generated stoploss, with the purpose of making it's win/loss ratio better comparable between strategies by using the same risk/reward everytime. 

setting it at "break even" once 100% risk/reward is reached, seems to be a good way to still give enough wiggle room while still minimizing risk. once the take profit point (1:1.5, 1:2, 1:3, etc) is reached, the profit is locked in. therefore I don't care so much, if this is close enough to price, that maybe a stoploss would be triggered.

main purpose of this `custom_stoploss()` should be, that you can compare different strategies win/loss rate with the same kind of risk/reward ratio.

In a real strategy I would try to dynamically adjust the stoploss once take profit is reached (like you recommended, using ATR etc again) and implementing another sell/exit signal besides stoploss. (that's what I hoped I could use `sell_profit_offset` for, now I kind of have to do that in `custom_stoploss` as well). but anyways, that's beside the point.

user_data/strategies/fixed_riskreward_loss.py
JoeSchr: > Do i get this right - once break-even-ratio is reached, you'll set stoploss to "break even" (sell at 0% profit)?
correct

user_data/strategies/fixed_riskreward_loss.py
xmatthias: would need to have a look - but it'd not give you what you want anyway - as `sell_profit_offset` is a global setting (applicable to all pairs) - so unless you only allow 1 parallel trade, whatever you set here would then apply to all future pairs too.

JoeSchr: True, let's just leave it out then

JoeSchr: I can also take out the 2 stages take profit scheme, if you think it takes away from simplicity as an example

xmatthias: i guess the rest is okish - we should just try to have "instructive" strategies like this not use things that don't work :laughing:

xmatthias (approved):
👍

JoeSchr:
Hurray 🙌


__

Typed on a mobile, please excuse any typing mistakes or short-spokeness.

PR #139, opened 2021-03-25
Adding the HLHB Strategy

wluciano:

## Summary
The HLHB ("Huck loves her bucks!") System simply aims to catch short-term forex trends.

Solve the issue: #___

## Quick strategy idea

# Step 1: What indicators are required?
EMA (5)
EMA (10)
RSI (10) - applied to the median price (HL/2)

# Step 2: Explain the Buy Strategy
When the 5 EMA crosses above the 10 EMA from underneath and the RSI crosses above the 50.0 mark from the bottom. *

# Step 3: Explain the Sell Strategy
When the 5 EMA crosses below the 10 EMA from the top and the RSI crosses below the 50.0 mark from the top. *

* Make sure that the RSI did cross 50.0 from the top or bottom and not just ranging tightly around the level.

# Step 4: Exit Rules
Use a 50-pip trailing stop and a 200-pip profit target. This increases the chances of the system riding longer trends

### DIFF
--- a/user_data/strategies/hlhb.py
+++ b/user_data/strategies/hlhb.py
@@ -32,10 +32,10 @@ class hlhb(IStrategy):
     stoploss = -0.3211,
 
     # Trailing stoploss
-    trailing_stop = "True",
-    trailing_stop_positive = 0.0117,
-    trailing_stop_positive_offset = 0.0186,
-    trailing_only_offset_is_reached = "True",
+    trailing_stop = True
+    trailing_stop_positive = 0.0117
+    trailing_stop_positive_offset = 0.0186
+    trailing_only_offset_is_reached = True
 
     # Optimal timeframe for the strategy.
     timeframe = '4h'
--- a/user_data/strategies/hlhb.py
+++ b/user_data/strategies/hlhb.py
@@ -1,0 +1,128 @@
+# pragma pylint: disable=missing-docstring, invalid-name, pointless-string-statement
+
+import numpy as np  # noqa
+import pandas as pd  # noqa
+from pandas import DataFrame
+from freqtrade.strategy import IStrategy
+import talib.abstract as ta
+import freqtrade.vendor.qtpylib.indicators as qtpylib
+
+
+class hlhb(IStrategy):
+    """
+    The HLHB ("Huck loves her bucks!") System simply aims to catch short-term forex trends.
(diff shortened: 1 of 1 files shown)


user_data/strategies/hlhb.py
```diff
@@ -0,0 +1,128 @@
...
+
+class hlhb(IStrategy):
+    """
+    The HLHB ("Huck loves her bucks!") System simply aims to catch short-term forex trends.
+    More information in https://www.babypips.com/trading/forex-hlhb-system-explained
+    """
+
+    INTERFACE_VERSION = 2
+
+    position_stacking = "True",
```
xmatthias, line 19: This parameter is ... kinda not used / has no effect.

user_data/strategies/hlhb.py
```diff
@@ -0,0 +1,128 @@
...
+
+    # Optimal stoploss designed for the strategy.
+    # This attribute will be overridden if the config file contains "stoploss".
+    stoploss = -0.3211,
+
+    # Trailing stoploss
+    trailing_stop = "True",
+    trailing_stop_positive = 0.0117,
+    trailing_stop_positive_offset = 0.0186,
+    trailing_only_offset_is_reached = "True",
```
xmatthias, line 38: The trailing stop (syntax) you use here is not valid - they'll end up as tuples (because of the trailing comma) - which will cause problems and errors all over the place.

```suggestion
    trailing_stop = True
    trailing_stop_positive = 0.0117
    trailing_stop_positive_offset = 0.0186
    trailing_only_offset_is_reached = True
```

xmatthias (commented):
Just 2 minor hints to make sure there's no surprises when using the strategy as the syntax is slightly off

user_data/strategies/hlhb.py
wluciano: This came from here: https://www.freqtrade.io/en/latest/hyperopt/

> You can also enable position stacking in the configuration file by explicitly setting "position_stacking"=true.

xmatthias: yeah, but it's only working in the configuration - and only for backtesting, not for real runs :laughing:

xmatthias (approved):
LGTM :+1: 

Thanks for sharing!

PR #140, opened 2021-03-29
fix key not exist error

mablue:
some times dataframe indicators did not exist!
I use coinex data

## Summary

Solve the issue: #___

## Quick strategy idea

- <change log #1>
- <change log #2>

### DIFF
--- a/user_data/strategies/berlinguyinca/MultiRSI.py
+++ b/user_data/strategies/berlinguyinca/MultiRSI.py
@@ -1,6 +1,7 @@
 # --- Do not remove these libs ---
 from freqtrade.strategy.interface import IStrategy
 from pandas import DataFrame
+from numpy import nan
 # --------------------------------
 import talib.abstract as ta
 from technical.util import resample_to_interval, resampled_merge
@@ -35,6 +36,10 @@ class MultiRSI(IStrategy):
         # resample our dataframes
         dataframe_short = resample_to_interval(dataframe, self.get_ticker_indicator() * 2)
         dataframe_long = resample_to_interval(dataframe, self.get_ticker_indicator() * 8)
+        if 'resample_8_rsi' not in dataframe.index:
+            dataframe['resample_8_rsi'] = nan
+        if 'resample_2_rsi' not in dataframe.index:
+            dataframe['resample_2_rsi'] = nan
 
         # compute our RSI's
         dataframe_short['rsi'] = ta.RSI(dataframe_short, timeperiod=14)


user_data/strategies/berlinguyinca/MultiRSI.py
```diff
@@ -35,6 +36,10 @@ def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame
         # resample our dataframes
         dataframe_short = resample_to_interval(dataframe, self.get_ticker_indicator() * 2)
         dataframe_long = resample_to_interval(dataframe, self.get_ticker_indicator() * 8)
+        if 'resample_8_rsi' not in dataframe.index:
+            dataframe['resample_8_rsi'] = nan
```
xmatthias, line 40: I'm not sure this is necessary, unless you have no data / not enough candles (in which case you have bigger problems anyway).

The name is fully dynamic (used with the same dynamic in `populate_indicators` as well as `populate_buy` and `populate_sell` methods.

In this example strategy - timeframe is 5m, so the indicator column name will be 8 * 5 = 40.


So this is not the correct name for the column as it assumes a column of 8 / 2 ... which would only apply when using the 1m timeframe.

it is however fully dynamic, so using 40 here is wrong as well and will lead to the same problem once should you use a different timeframe.

obviously the same also applies to the below method.

xmatthias (changes requested):

user_data/strategies/berlinguyinca/MultiRSI.py
mablue: Heh!! I think my exchange(coinex) is not 
compatible with ccxt cuz in all of loops i just get 8
So i think my get_ticker_indicator result allways return 1 :||||
|
|
||
 :|

xmatthias: if you only get a history of 8 candles, it'll be almost impossible to calculate any indicators.

mablue: Maybe for example in 1h timeframe I have a crypto corrency that added to the exchange site in last 4 hours!
And his data not enough to calculate anything!
If you want to improve flexablity of a code you will test on a coinex like site and with a crasy config and crazy commands(like ctrl+z + manual change of orders) to undrestand maximalism of the shit that can happend.
Or no
Maybe I will come to the binance with an out country vps...or proxy! Yes its not hard:)

xmatthias: that's not the point (and i'm not saying it shouldn't be safety-checked) - but your checks are not working unless you change the timeframe.

mablue: Im in the binance babe :) bot works perfect! you working on a sexy softwere man!!!!

PR #149, opened 2021-04-05
Fix hlhb

Samaoo:
- Delete the commas so that freqtrade doesn't throw a fatal exception when backtesting


xmatthias (approved):
super thanks :+1:

PR #177, opened 2021-05-15
Dynamic DNA size detection

snussik:

## Summary
The comment `# /5: Cuz We have 5 Group of variables inside buy_param` is not correct as we don't allways have exactly 5 groups. And in some cases, if params number less then 5, the hardcoded 5 divider can give us int(0.8) = 0 & empty         conditions list.


xmatthias (approved):
LGTM :+1:

PR #164, opened 2021-04-26
hyperopt Algorythm fixed + ZEUS Strategy and hyp+ADXMomentum hyperopt + GodStraHoHo + GodStra SpeedUp + Fixed new correncys missing data error.(no need to agefilter now) + no need to ta library

mablue:
## Summary
Its an strategy for find best Multiple ma conditions mix for trading 

## Quick strategy idea
the strategy compare multiple simple moving averages to find best place to trading.
it will also check the previous candles to get better results

PR #180, opened 2021-05-18
Mab stra

mablue:

## Summary

Solve the issue: #___

## Quick strategy idea

- <change log #1>
- <change log #2>


user_data/strategies/mabStra.py
```diff
@@ -9,58 +13,54 @@
...
+    buy_div_max = DecimalParameter(0, 2, decimals=4, default=2.25446, space='buy')
+    buy_div_min = DecimalParameter(0, 2, decimals=4, default=0.29497, space='buy')
+    # sell params
+    sell_mojo_ma_timeframe = IntParameter(2, 100, default=7, space='sell')
+    sell_fast_ma_timeframe = IntParameter(2, 100, default=14, space='sell')
+    sell_slow_ma_timeframe = IntParameter(2, 100, default=28, space='sell')
+    sell_div_max = DecimalParameter(0, 2, decimals=4, default=1.54593, space='sell')
+    sell_div_min = DecimalParameter(0, 2, decimals=4, default=2.81436, space='sell')
+
+    stoploss = -1
```
xmatthias, line 31: I find this pretty dangerous, and would not want us to advertise running without stoploss - as this can result in a loss of 100% of the money in the account.
maybe change it to something reasonable (10-20% maybe?) ... ?

xmatthias (commented):
Looking at the commit history - this are incremental PR's - each containing all files from the previous PR's.

Ideally (and that's what i meant in #164 ) - you'd create a new branch per PR - and then add one strategy (or strategy/hyperopt combo) to that.
Before creating the next branch, you switch back to the master branch - so your base is master again.
Otherwise, the "last" PR will contain all files from the previous pull requests - which (while in theory it's not a problem) can lead to merge conflicts.

user_data/strategies/mabStra.py
mablue: Yes why not??!

xmatthias (approved):
Great!
Thanks for sharing your work! :+1:

PR #181, opened 2021-05-18
Multi ma

mablue:

## Summary

Solve the issue: #___

## Quick strategy idea

- <change log #1>
- <change log #2>


user_data/strategies/MultiMa.py
```diff
@@ -0,0 +1,90 @@
...
+        # SMA - Simple Moving Average
+
+        for i in range(1, self.buy_ma_count.value+1):
+            dataframe[f'buy-ma-{i}'] = ta.SMA(dataframe,
+                                              timeperiod=int(i * self.buy_ma_gap.value))
+
+        for i in range(1, self.sell_ma_count.value+1):
+            dataframe[f'sell-ma-{i}'] = ta.SMA(dataframe,
+                                               timeperiod=int(i * self.sell_ma_gap.value))
+        print(dataframe.keys())
```
xmatthias, line 54: You should definitely comment or remove this - it's rather verbose.

user_data/strategies/MultiMa.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+        # SMA - Simple Moving Average
+
+        for i in range(1, self.buy_ma_count.value+1):
+            dataframe[f'buy-ma-{i}'] = ta.SMA(dataframe,
+                                              timeperiod=int(i * self.buy_ma_gap.value))
+
+        for i in range(1, self.sell_ma_count.value+1):
+            dataframe[f'sell-ma-{i}'] = ta.SMA(dataframe,
```
xmatthias, line 52: This will not work the way you expect it to. 
using `self.buy_ma_gap.value` in populate_indicators` will only give you ONE value (the default0 during hyperopt (`populate_indicators` will only run once during hyperopt, not once per epoch).

You should also use `for i in self.<param>.range:` - that's the simpler way to build a range of indicators - and has different effects between backtesting and hyperopt as it will will only create the full range  during hyperopt, and use the "selected" value during backtesting / live - therefore optimizing backtesting speed.

[details](https://www.freqtrade.io/en/latest/hyperopt/#optimizing-an-indicator-parameter).

xmatthias (changes requested):

user_data/strategies/MultiMa.py
mablue: O sorry

user_data/strategies/MultiMa.py
mablue: Ops I will use .range instad to range() 
Than if it just give me one time run than I shoud move them inside populate buy and sell !!
O maybe I just will remove the timeframe hyperoption

xmatthias (approved):
:+1:

Issue #185, opened 2021-05-24
Cant hyperopt DevilStra


rodrigogs:
Not able to hyperopt with DevilStra strategy. It works for like 30 epochs, and then I always get a:
```python
 [Epoch 24 of 300 (  8%)] |████████████\                                                                                                                                                      | [ETA:   0:31:59, Elapsed Time: 0:02:46]
2021-05-23 20:55:45,774 - freqtrade - ERROR - Fatal exception!
Traceback (most recent call last):
  File "c:\users\rodri\ideaprojects\trade-stuff-public\freqtrade\freqtrade\main.py", line 37, in main
    return_code = args['func'](args)
  File "c:\users\rodri\ideaprojects\trade-stuff-public\freqtrade\freqtrade\commands\optimize_commands.py", line 86, in start_hyperopt
    hyperopt.start()
  File "c:\users\rodri\ideaprojects\trade-stuff-public\freqtrade\freqtrade\optimize\hyperopt.py", line 429, in start
    asked = self.opt.ask(n_points=current_jobs)
  File "c:\users\rodri\anaconda3\envs\freqtrade\lib\site-packages\skopt\optimizer\optimizer.py", line 395, in ask
    x = opt.ask()
  File "c:\users\rodri\anaconda3\envs\freqtrade\lib\site-packages\skopt\optimizer\optimizer.py", line 367, in ask
    return self._ask()
  File "c:\users\rodri\anaconda3\envs\freqtrade\lib\site-packages\skopt\optimizer\optimizer.py", line 447, in _ask
    for xi in self.Xi])
  File "c:\users\rodri\anaconda3\envs\freqtrade\lib\site-packages\skopt\optimizer\optimizer.py", line 447, in <listcomp>
    for xi in self.Xi])
  File "c:\users\rodri\anaconda3\envs\freqtrade\lib\site-packages\skopt\space\space.py", line 1132, in distance
    distance += dim.distance(a, b)
  File "c:\users\rodri\anaconda3\envs\freqtrade\lib\site-packages\skopt\space\space.py", line 737, in distance
    if not (a in self and b in self):
  File "c:\users\rodri\anaconda3\envs\freqtrade\lib\site-packages\skopt\space\space.py", line 714, in __contains__
    return point in self.categories
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

xmatthias:
to explain what happens:
[...]

PR #188, opened 2021-05-25
fixing #185

mablue:


user_data/strategies/DevilStra.py
```diff
@@ -565,7 +576,7 @@ def populate_buy_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
 
         pairs = self.dp.current_whitelist()
         pair_index = pairs.index(metadata['pair'])
-        buy_params_index = self.buy_spell.value[pair_index]
+        buy_params_index = self.buy_spell.value.split(",")[pair_index]
```
xmatthias, line 579: After realizing how this strategy works (one different "spell" per pair) - i think you'll want to put a disclaimer / instruction at the top which points out that users MUST use a static pairlist - and MUST rehyperopt should that pairlist ever change.


This part (and the sell counterpart) is somewhat dangerous and can lead to exceptions - assuming i hyperopt with 20 pairs - and then run with 30 pairs - i'd get an `IndexError`.

Now this is not new to this PR but was there before (but i overlooked that part).

I'd rather see this handled gracefully (check if the split value is of the proper length).

mablue: yes its just useable for static pairlist and also i will note that on the top and also I will find a way to check compatiblity of indexes between two . thanks for help

user_data/strategies/DevilStra.py
```diff
@@ -527,45 +534,78 @@ def condition_generator(dataframe, operator, indicator, crossed_indicator, real_
...
     }
+
+    # Stoploss:
+    stoploss = -0.301
+
+    # Trailing stop:
+    trailing_stop = True
+    trailing_stop_positive = 0.01
+    trailing_stop_positive_offset = 0.027
+    trailing_only_offset_is_reached = True
```
xmatthias, line 565: On 1h candles, these settings will eliminate the ROI settings, as a move of 1% per candle is very likely on most pairs - therefore it'll almost always sell at "high - 1%" - resulting in inflated (and missleading) backtest results.

mablue: yes with just roi it can more profitable
but the hyperopt algo generate this things.
It can also generate more profitable parameters without roi as well!
let me to remove this things and re hype

xmatthias (approved):
:+1:

PR #196, opened 2021-06-05
wtc strategy added

mablue:
wtc

WTC Strategy: WTC(World Trade Center Tabriz)
is the biggest skyscraper of Tabriz, city of Iran
(What you want?it not enough for you?that's just it!)
No, no, I'm kidding. It's also mean Wave Trend with Crosses
algo by LazyBare(in TradingView) that I reduce it
signals noise with dividing it to Stoch-RSI indicator.
Also thanks from discord: @aurax for his/him
request to making this strategy.
hope you enjoy and get profit
# IMPORTANT: install sklearn befoure you run this strategy:
pip install sklearn
freqtrade hyperopt --hyperopt-loss SharpeHyperOptLoss --spaces buy sell --strategy wtc

### DIFF
--- a/user_data/strategies/wtc.py
+++ b/user_data/strategies/wtc.py
@@ -1,0 +1,153 @@
+# WTC Strategy: WTC(World Trade Center Tabriz)
+# is the biggest skyscraper of Tabriz, city of Iran
+# (What you want?it not enough for you?that's just it!)
+# No, no, I'm kidding. It's also mean Wave Trend with Crosses
+# algo by LazyBare(in TradingView) that I reduce it
+# signals noise with dividing it to Stoch-RSI indicator.
+# Also thanks from discord: @aurax for his/him
+# request to making this strategy.
+# hope you enjoy and get profit
+# Author: @Mablue (Masoud Azizi)
+# IMPORTANT: install sklearn befoure you run this strategy:
+# pip install sklearn
+# github: https://github.com/mablue/
+# freqtrade hyperopt --hyperopt-loss SharpeHyperOptLoss --spaces buy sell --strategy wtc
+
+from freqtrade.strategy.hyper import DecimalParameter
+import freqtrade.vendor.qtpylib.indicators as qtpylib
+import talib.abstract as ta
+from freqtrade.strategy import DecimalParameter
+from freqtrade.strategy import IStrategy
+from pandas import DataFrame
+from datetime import datetime
+#
+# --- Do not remove these libs ---
+import numpy as np  # noqa
+import pandas as pd  # noqa
+from sklearn import preprocessing
+
+# --------------------------------
+# Add your lib to import here
+
+
+class wtc(IStrategy):
+    ################################ SETTINGS ################################
+    # 61 trades. 16/0/45 Wins/Draws/Losses.
+    # * Avg profit: 132.53%.
+    # Median profit: -12.97%.
+    # Total profit: 0.80921449 BTC ( 809.21Σ%).
+    # Avg duration 4 days, 7:47:00 min.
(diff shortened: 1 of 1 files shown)


user_data/strategies/wtc.py
```diff
@@ -0,0 +1,153 @@
...
+            wt2 = ta.SMA(np.nan_to_num(wt1), 4)
+
+            dataframe['wt1'], dataframe['wt2'] = wt1, wt2
+
+            stoch = ta.STOCH(dataframe, 14)
+            slowk = stoch['slowk']
+            dataframe['slowk'] = slowk
+
+            x = dataframe.iloc[:, 1:].values  # returns a numpy array
+            min_max_scaler = preprocessing.MinMaxScaler()
```
xmatthias, line 117: That's a no-go.
If you run backtesting with this, it'll use the full data range for scaling - so it's implicitly creating a lookahead bias by using the full data.

Think about a pair that had big raises (say BTC) - and you backtest for 2 years. The first year will be scaled to 0.0-0.03 - while the remaining might be scaled to 0.03-1.0 ... as such - it's a lookahead as the scaler knows what the maximum (in the future!) has been.

xmatthias (changes requested):

user_data/strategies/wtc.py
mablue: ops! than I think I will
normalize just oscillators.;) sorry

xmatthias: Just normalizing the oscillators would work - but it's not what you're doing. 
`wt1` and `wt2` are not oscillators - but reflect the price - and as such - you cannot scale them to be reliably between 0 and 1 (because that will always require you to know the final value - which you can't as you're then doing lookahead.

mablue: yes maybe them are not oscillators I seen befour examples that go down and down and still down more and more!
but if you add this indicator on tradingview you can see its halfly works like an oscillators.and yes just normalizing is ok for making profit.check it:
https://www.tradingview.com/script/jFQn4jYZ-WaveTrend-with-Crosses-LazyBear/

mablue: Is there still other problems?!

mablue: Screenshot_20210609-000348

user_data/strategies/wtc.py
```diff
@@ -0,0 +1,153 @@
...
+# Also thanks from discord: @aurax for his/him
+# request to making this strategy.
+# hope you enjoy and get profit
+# Author: @Mablue (Masoud Azizi)
+# IMPORTANT: install sklearn befoure you run this strategy:
+# pip install sklearn
+# github: https://github.com/mablue/
+# freqtrade hyperopt --hyperopt-loss SharpeHyperOptLoss --spaces buy sell --strategy wtc
+
+from freqtrade.strategy.hyper import DecimalParameter
```
xmatthias, line 16: You should remove this import - it's imported twice (and this is the wrong, undocumented import, so internal freqtrade changes might break this line).

mablue: oh sorry I dont know why vscode add this things without asking me :( i will disable auto imports but its not bad sometimes to use.
vscode is growing like a deracula and didnt see his mistakes

xmatthias: vscode usually imports from the "direct" path. While that's not per se wrong - (and vscode shouldn't add this import if you had it already there before) - it might lead to problems down the road (we export things form the .strategy package intentionally - to avoid having users import from everywhere in the package, which would hinder refactorings.

xmatthias (approved):
:+1:

PR #197, opened 2021-06-06
update notice

mablue:

## Summary

Solve the issue: #___

## Quick strategy idea

- <change log #1>
- <change log #2>

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,6 +1,7 @@
 *For requestion a new strategy. Please use the template below.*  
 *Any strategy request that does not follow the template will be closed.*
-
+## Step 0: Are you switched specific branch of freqtrade and updated it and rechecked the error? 
+* If not solved than go to the next step
 ## Step 1: What indicators are required?
 *Please list all the indicators required for the buy and sell strategy.*
 


xmatthias:
I don't think that's a good way to do this - as the template is really made for new strategy requests - not so much for problem reports with strategies (we never got these in the past, really, up to your strategies :laughing: )

I'll think of something - but it'll probably be split templates for requests and problem reports (similar to the main repo setup).

mablue:
> I don't think that's a good way to do this - as the template is really made for new strategy requests - not so much for problem reports with strategies (we never got these in the past, really, up to your strategies 😆 )
> 
> I'll think of something - but it'll probably be split templates for requests and problem reports (similar to the main repo setup).

yes sometimes my strategys finds some bugs on main librarys like qtpylib and scikit-optimize too 😆 .
but in full look to the problem cuz of that freqtrade is verry innovative and fast updated repo and peoples sometimes take mistake about updating i get some this like problems on other networks and I trying to find the problem remotly but after days we undrestand that the user freqtrade is old :))

xmatthias:
Yeah - i've now added different templates for issues (issue request, issue bug) - with the strategy-bug requesting the same information as the main repository does.

Issue #30, opened 2019-05-08
SuperTrend 


bulentongun:
*For requestion a new strategy. Please use the template below.*  
*Any strategy request that does not follow the template will be closed.*

I didn't manage to add Supertrend to Freqtrade, and need help if it's possible. 

## Step 1: What indicators are required?
SuperTrend
F (factor): 2.1
Pd (period): 8

## Step 2: Explain the Buy Strategy
When the price is over the trend then BUY

## Step 1: Explain the Sell Strategy
When the price is below the trend then SELL

## Source
What come from this strategy? Cite your source:
https://www.tradingview.com/script/sufFpGjC-Supertrend-V1-0-Buy-or-Sell-Signal/
https://github.com/arkochhar/Technical-Indicators/blob/master/indicator/indicators.py

bulentongun:
Also ,
i have found this, 

`def ATR(df, period, ohlc=['Open', 'High', 'Low', 'Close']):
    """
    Function to compute Average True Range (ATR)
    
    Args :
        df : Pandas DataFrame which contains ['date', 'open', 'high', 'low', 'close', 'volume'] columns
        period : Integer indicates the period of computation in terms of number of candles
        ohlc: List defining OHLC Column names (default ['Open', 'High', 'Low', 'Close'])
        
    Returns :
        df : Pandas DataFrame with new columns added for 
            True Range (TR)
            ATR (ATR_$period)
    """
    atr = 'ATR_' + str(period)

    # Compute true range only if it is not computed and stored earlier in the df
    if not 'TR' in df.columns:
        df['h-l'] = df[ohlc[1]] - df[ohlc[2]]
        df['h-yc'] = abs(df[ohlc[1]] - df[ohlc[3]].shift())
        df['l-yc'] = abs(df[ohlc[2]] - df[ohlc[3]].shift())
         
        df['TR'] = df[['h-l', 'h-yc', 'l-yc']].max(axis=1)
         
        df.drop(['h-l', 'h-yc', 'l-yc'], inplace=True, axis=1)

    # Compute EMA of true range using ATR formula after ignoring first row
    EMA(df, 'TR', atr, period, alpha=True)
    
    return df

def SuperTrend(df, period, multiplier, ohlc=['Open', 'High', 'Low', 'Close']):
    """
    Function to compute SuperTrend
    
    Args :
        df : Pandas DataFrame which contains ['date', 'open', 'high', 'low', 'close', 'volume'] columns
[...]

PR #200, opened 2021-06-21
Includes Supertrend strategy

juankysoriano:
## Summary
As seen on issue #30 some users where requesting an approach for `Supertrend` strategy.

image

Solve the issue: #30

## Quick strategy idea

The proposed approach uses 3 super trend indicators with different values for `multiplier` and `period` parameters:
`[1, 10], [2, 11], [3, 12]`. Please note that additional work would be required in order to optimize these parameters, either via manual tuning or using `hyperopt`. The values for the parameters on this proposal were taken from a random youtube video.

# NOTE

I am very new to `freqtrade` so also opening the PR with the intention of learning a bit. Please feel free to take a look to the proposed strategy and suggest what I could be potentially be doing wrong here. I am not looking for strategy optimization, that's not my goal, but for learning about how to implement strategies. 

If the proposed strategy is wrongly implemented because my lack of knowledge this is where I would encourage readers to contribute 🙏


user_data/strategies/Supertrend.py
```diff
@@ -0,0 +1,115 @@
+"""
+Supertrend strategy:
+* Description: Generate a 3 supertrend indicators based on different pairs for [period, multiplier] -> [period: 10, multiplier: 1], [period: 11, multiplier: 2], [period: 13, multiplier: 3]
+               Buys if the 3 indicators are 'up'
+               Sells if the 3 indicators are 'down'
+* Author: @juankysoriano (Juan Carlos Soriano)
+* github: https://github.com/juankysoriano/
+"""
+
+# IMPORTANT: INSTALL TA BEFOUR RUN(pip install ta)
```
xmatthias, line 10: that's just "stolen" from somewhere - your strategy doesn't really depend on ta ... :laughing:

user_data/strategies/Supertrend.py
```diff
@@ -0,0 +1,115 @@
...
+            'sell'] = 1
+
+        return dataframe
+
+
+    """
+        Supertrend Indicator; adapted for freqtrade
+        from: https://github.com/freqtrade/freqtrade-strategies/issues/30
+    """
+    def supertrend(self, dataframe: DataFrame, multiplier, period):
```
xmatthias, line 77: i'd be cautious with supertrend - and amke sure to plot it and compare to reference charts.

that issue contains about 3 or 4 slightly different implementations - and i'm neither sure which one you picked, nor which one would be the "most correct".

In general, i'd think an indicator should be added to [technical](https://github.com/freqtrade/technical) - otherwise it's exclusive to this strategy (and an eventually discovered bug or incorrectness in that indicator would have to be fixed in every strategy that implements it).

user_data/strategies/Supertrend.py
```diff
@@ -0,0 +1,115 @@
...
+    trailing_only_offset_is_reached = False
+
+    timeframe = '1h'
+
+    startup_candle_count = 10
+
+    def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+        dataframe['supertrend_1'] = self.supertrend(dataframe, 1, 10)['STX']
+        dataframe['supertrend_2'] = self.supertrend(dataframe, 2, 11)['STX']
+        dataframe['supertrend_3'] = self.supertrend(dataframe, 3, 12)['STX']
```
xmatthias, line 44: As you correctly mention - the parameters could be pretty easily [hyperopted by using a parameter](https://www.freqtrade.io/en/latest/hyperopt/#optimizing-an-indicator-parameter).

That said - i don't see having every strategy parametrizable as a requirement - so it's up to you (the link is just a suggestion in case you're interrested.

user_data/strategies/Supertrend.py
```diff
@@ -0,0 +1,115 @@
...
+    # ROI table:
+    minimal_roi = {
+        "0": 0.087,
+        "372": 0.058,
+        "861": 0.029,
+        "2221": 0
+    }
+
+    # Stoploss:
+    stoploss = -0.265
```
xmatthias, line 29: in general, i'm sceptical of every strategy that uses a stoploss > 10/15%.

it'll look good on paper - but drawdown in case of a crash is usually pretty high.
again - that's more a "personal tip" than something that "must change" ... :smile:

user_data/strategies/Supertrend.py
juankysoriano: You are right I stole it 😅 and you are right, it is not needed. 

I am super newbie to this (and python as well) so I got confused by this import required by the `supertrend` function:

```
import talib.abstract as ta
```

Looks like `talib` and the `ta` I mention to install is not the same lib! will change

user_data/strategies/Supertrend.py
juankysoriano: Yes, this is very interesting. 

What would be the best way to compare this indicator implementation against reference charts? Is this one of the scenarios where plotting would be useful? Or are there other approaches typically taken? I am eager to know more.

In terms of defining the indicator on technical you are totally right. I am afraid that would require further knowledge on my side, on both python and freqtrade (I am thinking here about testing the indicator). 

@xmatthias I assume just moving this `supertrend` indicator to that repo is not what is actually expected, I can imagine I'd have to provide some tests and validation on the indicator for the community to start using it. So, here is my question: What has to be true in order to have this `indicator` merged into `technical`? A checklist would be appreciated.

user_data/strategies/Supertrend.py
juankysoriano: to be fair, I gave it a try, but failed, it's still on my personal roadmap to dig further on the `hyperopt` topic so I don't discards future contributions on this same strategy on that regard.

user_data/strategies/Supertrend.py
juankysoriano: On this case the value was suggested by `hyperopt` the `stoploss` and `trail` spaces. In general I am seeing very high values for `stoploss` when doing `hyperopt` on the space. 

What would be the best take here, from a strategy planning perspective? Taking the resutls from`hyperopt` with a grain of salt and considering lower values on `live` cases?

user_data/strategies/Supertrend.py
juankysoriano: resolved on d705981

user_data/strategies/Supertrend.py
xmatthias: in the end, stoploss is personal preference - thee's people that run without stoploss - personally, i don't hyperopt stoploss but instead set it based on my risk expectations.

stoploss optimization will slightly depend on the loss-function that's used - but it'll often favor higher stoplosses - which is a good strategy in bull markets, but will fail in bear markets ...

user_data/strategies/Supertrend.py
xmatthias: Reference is difficult - many seem to tread tradingview as "THE" reference - to which i'd disagree.
Many of their (builtin) calculations are closed-source, and often slightly deviate from the original papers (we've got quite some examples of this distributed across the freqtrade repository issues).

As such - i always "try" to find the original paper and refer to that implementation.

Regarding technical - there SHOULD be tests - but in many cases there are none (it's a library we inherited - and writing tests fo this takes a lot of time).

It's also difficult to determine WHAT to test - 
is it the result numbers (all indicators are usually self-contained, so changes will always be done explicitly)? 
or simply the result format (so the return value is a dataframe with 2 columns named X and Y)?

juankysoriano: > As such - i always "try" to find the original paper and refer to that implementation.

I will try to give this a try, but on the core topic I totally agree, we need to make sure the `supertrend` function is totally accurate in terms of implementation if we want to encourage others to use this strategy, and for that we need to have it properly implemented on `technical`. 

At the moment, as this is, I see this as a blocker for merging this PR for which what I would suggest in case of you agreeing is that I rename this strategy to something else (something along the lines of `ApproachToSupertrend` and add a big comment on the `supertrend` method  (also renaming it) alerting others to take it with a grain of salt, being it not an official implemetation backed by `freqtrade`. 

Otherwise I'd have to close the PR until I manage to - or someone else does - implement the indicator on `technical`

user_data/strategies/Supertrend.py
juankysoriano: > stoploss optimization will slightly depend on the loss-function that's used - but it'll often favor higher stoplosses - which is a good strategy in bull markets, but will fail in bear markets ...

Yes, at the end there is some degree of overfitting with `hyperopt`,  if the market never crashed on the timerange you hyperopt for, it's likely it would suggest a large `stoploss` which puts the trader on risk for potential market crashes.

Thanks for the tip 🤝

juankysoriano: Added a comment for the readers f34101d

user_data/strategies/Supertrend.py
juankysoriano: Ah, just a note, the `supertrend` here is the one that @synackSA is suggesting here to be matching the Binance's integration of Tradingview's chart: https://www.tradingview.com/v/r6dAP7yi/

I understand it's still an ongoing conversation, so we'll have to understand what `supertrend` indicator implementation we want to use before moving this to `technical`. 

I don't really know how to proceed here, but for the time being either we keep the PR open until we find a resolution on what `supertrend` implementation is proposed to `technical` or I just simply rename this strategy to avoid others thinking that this is some sort of official implementation for `supertrend`

user_data/strategies/Supertrend.py
juankysoriano: Something I don't fully understand about `hyperopt` for parameter optimization is the following:

Say that I define 6 parameters (3 are for the parameters on indicators that will be used on the `buy` space; 3 por the parameters on indicators that will be used for the `sell` space). (I understand there is no combined space for `sell` and `buy`, so it has to be 6 in total if I want to use 3 indicators for super trend).

```
buy_m1 = IntParameter(1, 7, default=4)
    buy_m2 = IntParameter(1, 7, default=4)
    buy_m3 = IntParameter(1, 7, default=4)
    buy_p1 = IntParameter(7, 21, default=14)
    buy_p2 = IntParameter(7, 21, default=14)
    buy_p3 = IntParameter(7, 21, default=14)

    sell_m1 = IntParameter(1, 7, default=4)
    sell_m2 = IntParameter(1, 7, default=4)
    sell_m3 = IntParameter(1, 7, default=4)
    sell_p1 = IntParameter(7, 21, default=14)
    sell_p2 = IntParameter(7, 21, default=14)
    sell_p3 = IntParameter(7, 21, default=14)
```

Now, following the reading you suggest I would be using them like this:

```

    def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
        for multiplier in self.buy_m1.range:
            for period in self.buy_p1.range:
                dataframe[f'supertrend_1_buy_{multiplier}_{period}'] = self.supertrend(dataframe, multiplier, period)['STX']

        for multiplier in self.buy_m2.range:
            for period in self.buy_p2.range:
                dataframe[f'supertrend_2_buy_{multiplier}_{period}'] = self.supertrend(dataframe, multiplier, period)['STX']

        for multiplier in self.buy_m3.range:
            for period in self.buy_p3.range:
                dataframe[f'supertrend_3_buy_{multiplier}_{period}'] = self.supertrend(dataframe, multiplier, period)['STX']

        for multiplier in self.sell_m1.range:
            for period in self.sell_p1.range:
                dataframe[f'supertrend_1_sell_{multiplier}_{period}'] = self.supertrend(dataframe, multiplier, period)['STX']

        for multiplier in self.sell_m2.range:
            for period in self.sell_p2.range:
                dataframe[f'supertrend_2_sell_{multiplier}_{period}'] = self.supertrend(dataframe, multiplier, period)['STX']

        for multiplier in self.sell_m3.range:
            for period in self.sell_p3.range:
                dataframe[f'supertrend_3_sell_{multiplier}_{period}'] = self.supertrend(dataframe, multiplier, period)['STX']

        return dataframe

        dataframe.loc[
            (
               (dataframe[f'supertrend_1_buy_{self.buy_m1.value}_{self.buy_p1.value}'] == 'up') &
               (dataframe[f'supertrend_2_buy_{self.buy_m2.value}_{self.buy_p2.value}'] == 'up') &
               (dataframe[f'supertrend_3_buy_{self.buy_m3.value}_{self.buy_p3.value}'] == 'up') & # The three indicators are 'up' for the current candle
               (dataframe['volume'] > 0) # There is at least some trading volume
	    ),
            'buy'] = 1

        return dataframe

    def populate_sell_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
        dataframe.loc[
            (
               (dataframe[f'supertrend_1_sell_{self.sell_m1.value}_{self.sell_p1.value}'] == 'down') &
               (dataframe[f'supertrend_2_sell_{self.sell_m2.value}_{self.sell_p2.value}'] == 'down') &
               (dataframe[f'supertrend_3_sell_{self.sell_m3.value}_{self.sell_p3.value}'] == 'down') & # The three indicators are 'down' for the current candle
               (dataframe['volume'] > 0) # There is at least some trading volume
            ),
            'sell'] = 1

        return dataframe
```

And this works, `hyperopt` will eventually output 

```
    # Buy hyperspace params:
    buy_params = {
        "buy_m1": 4,
        "buy_m2": 7,
        "buy_m3": 1,
        "buy_p1": 8,
        "buy_p2": 9,
        "buy_p3": 8,
    }

    # Sell hyperspace params:
    sell_params = {
        "sell_m1": 1,
        "sell_m2": 3,
        "sell_m3": 6,
        "sell_p1": 16,
        "sell_p2": 18,
        "sell_p3": 18,
    }
```

But I have noted the `calculating indicators` bit of the process to take a long time (it's understandable, there is a quite wide range for the parameters value, and lots of operations to be performed on `populate_indicators`. 

Now, how do I proceed with that information that `hyperopt` is outputting, wouldn't it be very bad in terms of performance if I keep the class with the `populate_*` methods I showed before? (for example in `back-testing` or `dry_run`).

Is that "setup" for `populate_indicators` something that can still be used on `backtesting` or `live` scenarios?
Or on those scenarios the values from: 

```
    # Buy hyperspace params:
    buy_params = {
        "buy_m1": 4,
        "buy_m2": 7,
        "buy_m3": 1,
        "buy_p1": 8,
        "buy_p2": 9,
        "buy_p3": 8,
    }

    # Sell hyperspace params:
    sell_params = {
        "sell_m1": 1,
        "sell_m2": 3,
        "sell_m3": 6,
        "sell_p1": 16,
        "sell_p2": 18,
        "sell_p3": 18,
    }
``` 
are used instead of 
```
    buy_m1 = IntParameter(1, 7, default=4)
    buy_m2 = IntParameter(1, 7, default=4)
    buy_m3 = IntParameter(1, 7, default=4)
    buy_p1 = IntParameter(7, 21, default=14)
    buy_p2 = IntParameter(7, 21, default=14)
    buy_p3 = IntParameter(7, 21, default=14)

    sell_m1 = IntParameter(1, 7, default=4)
    sell_m2 = IntParameter(1, 7, default=4)
    sell_m3 = IntParameter(1, 7, default=4)
    sell_p1 = IntParameter(7, 21, default=14)
    sell_p2 = IntParameter(7, 21, default=14)
    sell_p3 = IntParameter(7, 21, default=14)
```

So the effective range is `1` and there is therefore not that big impact in performance?

user_data/strategies/Supertrend.py
xmatthias: first of all, there's nothing wrong with leaving a PR hanging until issues are fixed (even if that's issues that need other things to happen first).

To me, a comment at the top of the strategy hinting at this should be enough - i'd not even rename the method as this is one implementation of supertrend. 
I'm not certain we can come up with "the" implementation for supertrend - as in #30 we've seen ~3-4 implementations already - where noone seems certain which one is the correct.

user_data/strategies/Supertrend.py
xmatthias: `buy_m1.range` (or every other parameter) will have a range of 1 when not in hyperopt mode (or if that parameter is excluded from hyperopting) - so the performance hit you see is only there for hyperopt - not for any of the other modes.

By pasting the result into the strategy (make sure indentation is correct) - the resulting parameters are then used by the strategy.

juankysoriano: 🙇 thanks for the explanation, I have done a change to consider hyperoptimization of the parameters, following what we discussed:

cccdd6d

user_data/strategies/Supertrend.py
juankysoriano: I have added a comment 

2a0775a

juankysoriano:
Thanks @xmatthias for your time on reviewing this, I have learned a few very interesting things. 

Note for the readers, the best way of learning is contributing, so don't hesitate

xmatthias (approved):
Great contribution - thanks a lot! :+1:

colinbouvry:
Hello,

Very nice SuperTrend !

The strategy buy and sell too many times :
image

Any idea to buy or to sell only if changement of state ?

Any idea to plot SuperTrend under freqTrade like tradingview ?

Thanks

PR #202, opened 2021-06-26
heracles in New API, Bugs fixed

### LINKED ISSUE #201: Questions about Heracles strategy and hyperopt
Hello all and @mablue,

I run that hyperopt [Heracles](https://github.com/freqtrade/freqtrade-strategies/blob/master/user_data/hyperopts/HeraclesHo.py) created by @mablue with [the appropriate strategy](https://github.com/freqtrade/freqtrade-strategies/blob/master/user_data/strategies/Heracles.py).

It's running very well and the results of the hyperopt are amazing. I got this parameters:

```
# Buy hyperspace params:
buy_params = {
    "buy-div": 0.51844,
    "DFINDShift": 4,
    "DFCRSShift": 1,
}

# Sell hyperspace params:
sell_params = {
    "sell-rtol": 0.05468,
    "sell-atol": 0.00019,
    "DFINDShift": 4,
    "DFCRSShift": 1,
}
```

In the hyperopt file there is the following code in `populate_buy_trend`:
```
dataframe.loc[(DFIND < DFCRS), 'buy'] = 1
```

But in strategy file there is a different condition in `` for generating a buy signal:
```
conditions.append(
  DFIND.shift(params['DFINDShift']).div(DFCRS.shift(params['DFCRSShift'])) <= params['buy-div'])
```

@mablue why they are different? I can't put the hyperopt results in the strategy file because they are different.

What I tried:

I copied the code for signal generation from the hyperopt to the strategy file with parameters from the hyperopt file:
```

        IND = self.buy_params['buy-indicator-0']
        CRS = self.buy_params['buy-cross-0']
        DFIND = dataframe[IND]
        DFCRS = dataframe[CRS]

        dataframe.loc[
            (
                DFIND.shift(4).div(DFCRS.shift(1) <= 0.51844)
            ),
            'buy'] = 1

        return dataframe
```

```
def populate_sell_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
        IND = self.sell_params['sell-indicator-0']
        CRS = self.sell_params['sell-cross-0']

        DFIND = dataframe[IND]
        DFCRS = dataframe[CRS]

        dataframe.loc[
            (
                np.isclose(
                    DFIND.shift(4),
                    DFCRS.shift(1),
                    rtol=0.05468,
                    atol=0.00019
                )
            ),
            'sell'] = 1

        return dataframe
```

But if I try to run a backtest I get this error:
```
2021-06-24 14:42:38,308 - freqtrade - ERROR - Fatal exception!
Traceback (most recent call last):
  File "/freqtrade/freqtrade/main.py", line 37, in main
    return_code = args['func'](args)
  File "/freqtrade/freqtrade/commands/optimize_commands.py", line 53, in start_backtesting
    backtesting.start()
  File "/freqtrade/freqtrade/optimize/backtesting.py", line 517, in start
    min_date, max_date = self.backtest_one_strategy(strat, data, timerange)
  File "/freqtrade/freqtrade/optimize/backtesting.py", line 489, in backtest_one_strategy
    results = self.backtest(
  File "/freqtrade/freqtrade/optimize/backtesting.py", line 376, in backtest
    data: Dict = self._get_ohlcv_as_lists(processed)
  File "/freqtrade/freqtrade/optimize/backtesting.py", line 200, in _get_ohlcv_as_lists
    self.strategy.advise_buy(pair_data, {'pair': pair}), {'pair': pair})[headers].copy()
  File "/freqtrade/freqtrade/strategy/interface.py", line 764, in advise_buy
    return self.populate_buy_trend(dataframe, metadata)
  File "/freqtrade/user_data/strategies/Dobby.py", line 150, in populate_buy_trend
    dataframe.loc[
  File "/home/ftuser/.local/lib/python3.9/site-packages/pandas/core/indexing.py", line 688, in __setitem__
    indexer = self._get_setitem_indexer(key)
  File "/home/ftuser/.local/lib/python3.9/site-packages/pandas/core/indexing.py", line 630, in _get_setitem_indexer
    return self._convert_tuple(key, is_setter=True)
  File "/home/ftuser/.local/lib/python3.9/site-packages/pandas/core/indexing.py", line 754, in _convert_tuple
    idx = self._convert_to_indexer(k, axis=i, is_setter=is_setter)
  File "/home/ftuser/.local/lib/python3.9/site-packages/pandas/core/indexing.py", line 1212, in _convert_to_indexer
    return self._get_listlike_indexer(key, axis, raise_missing=True)[1]
  File "/home/ftuser/.local/lib/python3.9/site-packages/pandas/core/indexing.py", line 1266, in _get_listlike_indexer
    self._validate_read_indexer(keyarr, indexer, axis, raise_missing=raise_missing)
  File "/home/ftuser/.local/lib/python3.9/site-packages/pandas/core/indexing.py", line 1308, in _validate_read_indexer
    raise KeyError(f"None of [{key}] are in the [{axis_name}]")
KeyError: "None of [Float64Index([nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,\n              ...\n              inf, inf, inf, inf, inf, inf, inf, inf, inf, inf],\n             dtype='float64', length=343)] are in the [index]"
```

I run backtest with Docker-Compose:
```
version: "3.9"
services:
  BACKTEST:
    image: freqtrade_custom
    container_name: BACKTEST
    volumes: 
      - "./user_data:/freqtrade/user_data"
    command: >
      backtesting
      --export=trades 
      --export-filename=/freqtrade/user_data/backtest_results/backtest-dobby.json
      --config /freqtrade/user_data/conf-ho-dobby.json
      --timerange=20210101-20210621
      --strategy Dobby
      --fee 0.0015
      --logfile /freqtrade/user_data/logs/backtest.log
```

The custom image I create with:
```
version: "3.9"
services:
  CUSTOMIMAGE:
    image: freqtrade_custom
    build:
      context: .
      dockerfile: "./Dockerfile.ta"
```

In my `Dockerfile.ta` I have this code:
```
FROM freqtradeorg/freqtrade:latest
RUN pip install ta
```

@mablue I wonder if it is a `pandas` or strategy error?

Kind regards
@OtenMoten 

PS: I added the changed strategy file and my config as .TXT files.

[conf-ho-dobby.txt](https://github.com/freqtrade/freqtrade-strategies/files/6710236/conf-ho-dobby.txt)
[Dobby.txt](https://github.com/freqtrade/freqtrade-strategies/files/6710237/Dobby.txt)

mablue:
heracless strategy/hyperopt upgraded to New API
bugs fixed #201


user_data/strategies/Heracles.py
```diff
@@ -31,103 +26,154 @@
 import numpy as np
 
 
+def normalize(df):
+    # To enable normalization outcomment below line:
+    df = (df-df.min())/(df.max()-df.min())
```
xmatthias, line 31: You'll need to be careful with such approaches.

I understand you try to normalize this - but this will give totally different results between backtesting and live as it's looking at the full dataframe (and as such - is look-ahead biased / using data it should not).

user_data/strategies/Heracles.py
```diff
@@ -31,103 +26,154 @@
...
+
+
+    # buy params
+    buy_div = DecimalParameter(-5, 5, default=0.51844, decimals=4, space='buy')
+    buy_indicator_shift = IntParameter(-5, 5, default=4, space='buy')
+    buy_crossed_indicator_shift = IntParameter(-5, 5, default=1, space='buy')
+
+    # sell params
+    sell_rtol = DecimalParameter(1.e-10, 1.e-0, default=0.05468, decimals=10, space='sell')
+    sell_atol = DecimalParameter(1.e-16, 1.e-0, default=0.00019, decimals=10, space='sell')
```
xmatthias, line 75: I find this space overly large.
You'll not find any (added) value after 4 or 5  decimals - so using decimals=10 will inflate the search-space (slow down hyperoptimization) without providing any real benefit.

user_data/strategies/Heracles.py
mablue: yee this normalization is very usefull but just in oscilators not in indicators like sma or etc that have not a standard min & max.
also if you mean full dataframe : all columns. not it just 4 indicators that normalised that all of them has not following price and have min/max as well.
and we have not the problems like befour for this. see lines 87 and etc..

user_data/strategies/Heracles.py
mablue: yes I will fix it.i just do that cuz of having some coins that have very small prices like trtl/btc= 0.0000000000001 but ye we there just have indicators and no need to price comparison

user_data/strategies/Heracles.py
xmatthias: unfortunately, that's not true.
Assuming an oscillator like RSI (ranges from 0 to 100) for simplicity - it will apply to all other oscillators as well.

By scaling it to "dataframe's min/max" - you're scaling it to the dataframe's min/max values.
Assuming for backtesting - these will very likely be close to 100 and close to 0 (with a big enough time-range).
if you now do the same for only 500 candles each - it will heavily depend - you'll sometimes get it scaled between 40 and 100 ... sometimes between 30 and 60 ... 

you'll simply not know - but as you're using the scaled / normalized values - you'll work off of a wrong assumption.

If you only apply it to oscillators - then their absolute upper and lower limit must be known to you - so use that instead of using `.min()` and `.max()` (if they have unknown upper/lower limits, they're not really oscillators - they just look like oscillators).

xmatthias (changes requested):
Unfortunately, the normalization serves as a bad example - which can lead others to use it wrongly assuming it's something it isn't - and as such - please change that to something that either applies specifically to the indicator you're using - or to a method that does not suffer from the lookahead bias problem.

mablue:
> Unfortunately, the normalization serves as a bad example - which can lead others to use it wrongly assuming it's something it isn't - and as such - please change that to something that either applies specifically to the indicator you're using - or to a method that does not suffer from the lookahead bias problem / or rather (to be more precise) - to something that generates the same result for today, 10:00 - no matter if i'm using 500 candles - or 50000 candles (which currently, is simply not the case).

I can prementelly remove the normalization.there is no problem when I remove but I will change the limits of dividation values to maybe integer or check the devidation order to devide always smallest/biggest indicator to be allways inside 0~1 and set the limits of devidation to 0~1 and thats it we can use it without normalization.also I will rehype and repreduce defaults to decrese the danger of default usage.thank you for help to improve it ;)

user_data/strategies/Heracles.py
mablue: ops yes now I understand what you mean yes if we have just 10 candles in our hands its max and min rsi maybe will be 50~80 and this normalization will get us wrong values.
ok than I can find an optimal min/max to normalize things in it.
for example I can check 10 coins indicators and find the best value to set as min/max for each indicator and than calculate the normalized value.
also if we think honest, in this case our all normalizations in other strategies will preduce wrong values some times!I think I will change all normalization methods in other strategies too!!!

mablue: but I think the true way is that I just remove the normalization cuz some indicators have not a min or max and it make problems and also I cant really find all exact min or maxes in all genes in godStra to normalize,
or maybe the best technic to do is that we check the candles that we have and if its smaller than for example 50 we just ingnore it.
and also if a dataset have not 3 Repetitious min and max we will ignore it.and dont normalise it never!
i think the last tecnic will work well @xmatthias  yes?!

xmatthias: Looking at part of the dataframe can still cause the same effect (as you won't know which part you're looking at). 
To me, you can't properly normalize (ensure the values are between 1 and 0) unless you know the theoretical maxima's of the indicators - not without either looking into the future, or risking it to eventually flow over if there's some extreme values that didn't happen earlier).

As such (and this is why) i've always been of the opinion that blindly combining indicators without thinking twice (-> let hyperopt do that) is not a good approach.

> also if we think honest, in this case our all normalizations in other strategies will preduce wrong values some times

Not many strategies will use normalizations - and to be honest, i'm looking at your's a bit closer now - as they created more issues than any of the other strategies before :laughing:

mablue: maybe if I find a teorical min and max it will fixed but just in oscilators and some others maybe for example patterns that have just 0 and 100 values: but I think I will take care strategyis from unesessery normalizations.
for example in this stra we dont need really a normalization...
yes thank you to looking closer its take more time of you and I know. thanks.✌️😎✌️

user_data/strategies/Heracles.py
xmatthias: I find it interresting that you had this fixed (reduced to 4 decimals) and now put it back.

As mentioned before - using a 10 decimal spaces does not make sense and will overfit the strategy for sure.

mablue: opsss!! I dont know why so soory i will fix it

mablue:
OK.I changed the decimals also removed sell spaces cuz i think them are not useful and always roi was effective on sell

user_data/strategies/Heracles.py
mablue: @xmatthias OK.I changed the decimals also removed sell spaces cuz i think them are not useful and always roi was effective on sell

xmatthias (approved):
Great - looks good now :+1:

PR #204, opened 2021-07-14
Add new strategy to demonstrate custom_sell in Strategy001 example

froggleston:
## Summary
Add a very simple custom_sell example to Strategy001 to demonstrate

### DIFF
--- a/user_data/strategies/Strategy001_custom_sell.py
+++ b/user_data/strategies/Strategy001_custom_sell.py
@@ -1,0 +1,141 @@
+
+# --- Do not remove these libs ---
+from freqtrade.strategy.interface import IStrategy
+from typing import Dict, List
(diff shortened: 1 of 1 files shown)


xmatthias:
I understand that this is to serve an example (you can accomplish this type of sell also with settings).

Better question is - do we really need this (in a sample strategy people will be tempted to use "as is" ...) - i'm yet to see many questions on how to use it (usually the "how to get good parameters within the method" is the bigger question).

froggleston:
Then I would suggest we have strategies in there called "example-001" or something. The same arguments could be applied to all strats in that repo - are they all ready to use as is? No, but that doesn't mean the examples they provide are not useful.

xmatthias (approved):

PR #209, opened 2021-08-08
BreakEven.py -- sell everything at 0% or better profit

lazydroid:
## Summary

In case I want to stop the bot ASAP, this strategy sells everything at break even or better profit. A bit less radical version of `/forcesell all`, if you want.

## Quick strategy idea

Ignore sell / buy signals, set up `minimal_roi` to `{ 0 : 0 }` to make sure all open positions currently above the break even level are sold, wait for the rest to either rise or sink down to `stop_loss` level.

### DIFF
--- a/user_data/strategies/BreakEven.py
+++ b/user_data/strategies/BreakEven.py
@@ -43,7 +43,7 @@ class BreakEven(IStrategy):
 #    }
 
     # Optimal stoploss designed for the strategy
-    stoploss = -0.25
+    stoploss = -0.05
 
     # Optimal timeframe for the strategy
     timeframe = '5m'
--- a/user_data/strategies/BreakEven.py
+++ b/user_data/strategies/BreakEven.py
@@ -1,0 +1,67 @@
+# --- Do not remove these libs ---
+from freqtrade.strategy.interface import IStrategy
+from pandas import DataFrame
+# --------------------------------
+
+
+class BreakEven(IStrategy):
+    """
+    author@: lenik
+
+    Sometimes I want to close the bot ASAP, but not have the positions floating around.
+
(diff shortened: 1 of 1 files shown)


user_data/strategies/BreakEven.py
```diff
@@ -0,0 +1,67 @@
...
+#        "0": 0
+#    }
+
+    # And this is basically "/forcesell all", that sells no matter what profit
+#    minimal_roi = {
+#        "0": -1
+#    }
+
+    # Optimal stoploss designed for the strategy
+    stoploss = -0.25
```
xmatthias, line 46: You'll need to be careful with this.

assuming your original strategy is not using trailing-stop (or a trade's stoploss didn't start trailing yet) - then this would update the stoploss from an initial 5% value to 25% - which can often lead to bigger losses than anticipated (this will also counter the intend of this strategy, as you'll be stuck in trades a lot longer).

lazydroid: Sorry, had no idea the default is 5%, usually it's like 25-30% in most strats. Please, change to the appropriate level. Thanks!

xmatthias: there's no default per se - but to me, every strategy that uses a value > 10% as a bad strategy, as it's simply "sitting out losers" in hopes that they recover.

If you REALLY run into a 30% stoploss (or 2 or 3 in a row), it'll take ages (most often 10s of trades) to recover - as you'll not have 30% ROI before selling a trade, will you? and then your risk/reward ratio is way off.

lazydroid: I have no argument here, the original value came from the default strat. Changed to 5%.
Thank you for taking time to explain things!

xmatthias (approved):
LGTM :+1:

PR #213, opened 2021-08-17
Multima indexing problems fixed and DUALFIT added

mablue:
it had some indexing problem on calculating conditions that fixed also DUALFIT added to strategy


user_data/strategies/MultiMa.py
```diff
@@ -50,21 +91,20 @@ def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame
...
+        # exit()
         conditions = []
 
-        for i in self.buy_ma_count.range:
-            if i > 1:
-                shift = self.buy_ma_shift.value
-                for shift in self.buy_ma_shift.range:
+        for i in range(1, self.buy_ma_count.value+1):
+            dataframe[f'buy-ma-{i}'] = ta.EMA(dataframe,
+                                              timeperiod=int(i * self.buy_ma_gap.value))
```
xmatthias, line 100: I'm not a huge fan of calculating indicators in the `populate_*_trend()` methods.

You should rather use the [documented approach][https://www.freqtrade.io/en/stable/hyperopt/#optimizing-an-indicator-parameter) for this.

For hyperopt, this is pretty inefficient and will slow down hyperopt a lot, as this means that EMA1 (or whatever the first value is) will be calculated every single time - so you'll end up calculating the same exact EMA value over and over again (assuming other parameters changed).

By calculating the maximum range (say, from EMA1 to EMA100) in `populate_indicators` - it'll be calculated only once.

mablue: I used a new method to fix it. if there is calculated ema it will not calc again!
you can add it inside freqtrade too. I dont know how cuz user can say everything to his calculated indicator. and will his own check inside strategy
but if with a extra interpreter the ta.XXX words checked and than saved inside some file it will very helpfull for calculating non static(dynamic) indicator calculation inside buy and sell strategy..and it make hyperoption speedy.
but I think the only place that this things will needen are inside my strategies. no One uses dynamic calculating indicators nowadays...
But sometimes it use able...in hyperoption its very useable to calculate things right there inside the loop for finding best values  for ta methods

mablue: live checked
Screenshot from 2021-08-21 02-19-18

xmatthias: seems like the history of that branch is messed up.
Please checkout master and make sure it's uptodate with the repository
```
git checkout master
git fetch origin # (assuming origin is what you use for freqtrade/freqtrade-strategies - otherwise please adujust all usages of this below)
# Ensure origin 
git reset --hard origin/master 

git checkout <yourbranch>

# This might cause merge-conflicts - fix them, stage the files you fixed and move forward with `git rebase --continue`
git rebase master

git push -f
```

mablue: Im done some things but I dont know fixed or not?!!

user_data/strategies/MultiMa.py
```diff
@@ -13,109 +24,232 @@
...
+    # @formatter:on
+    # DONT TOUCH OR FORMAT ABOVE!
```
xmatthias, line 176: You might comment it this way - but i'l surely not merge a strategy in that has lines this long.
Python has conventions of "one command per line" on purpose.

this is neither understandable nor maintainable, and as such, cannot be accepted.

xmatthias (changes requested):

user_data/strategies/MultiMa.py
mablue: ok but I think you know why Im do this.
I'll fix it

mablue: > ok but I think you know why Im do this.for having clean code!
> I'll fix it

mablue:
@xmatthias multima will just use 3 ribbons for each  buy and sell. mass conditions fuckd up my mind :|

user_data/strategies/MultiMa.py
```diff
@@ -16,55 +19,95 @@
...
+    trailing_stop = False  # value loaded from strategy
+    trailing_stop_positive = None  # value loaded from strategy
+    trailing_stop_positive_offset = 0.0  # value loaded from strategy
+    trailing_only_offset_is_reached = False  # value loaded from strategy
+    
+    # Timeframe
+    timeframe = '5m'
+
+    
+    timeframes = [5, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100, 150, 200]
```
xmatthias, line 64: The naming of this variable (timeframes) suggests it's selecting different timeframes (5m, 30m, ...) - but it's actually selecting something completely different.

I think a better name will be appropriate.

user_data/strategies/MultiMa.py
```diff
@@ -1,10 +1,13 @@
 # MultiMa Strategy
+# MultiMa proposed to find a way with making Ribbon idea.
+# That can make signals by watching sort of Ribbon of sort of data.
```
xmatthias, line 3: That explanation does not really make sense.
You're kindof looping (i guess you mean that with ribbon, not a Ribbon on a package) - but what it tries to accomplish / compare is totally unclear.

mablue: this explanation is totally have english grammer problems. and changed and will commited. now I just try to get good signals in live. It have signal generation problems cuz of massive conditions.

user_data/strategies/Rainbow.py
```diff
@@ -0,0 +1,216 @@
...
+# And See What will happen. The key thing we use in this strategy Is Driver that controls
+# That Which (tf, rolling, shift) setting of an indicator will use to make Ribbons.
+# There isAcomperator param that used for find best comperatorare operator to generated ribbons.
+# freqtrade backtesting -s Rainbow -c configusdt.json
+# freqtrade hyperopt --hyperopt-loss ShortTradeDurHyperOptLoss --strategy Rainbow -c configusdt.json --eps
+# $ freqtrade hyperopt --hyperopt-loss OnlyProfitHyperOptLoss --strategy Rainbow -c configusdt.json --eps
+# $$$ freqtrade hyperopt --hyperopt-loss SharpeHyperOptLoss --strategy Rainbow -c configusdt.json --eps
+# $ freqtrade hyperopt --hyperopt-loss SharpeHyperOptLossDaily --strategy Rainbow -c configusdt.json --eps
+# $$ freqtrade hyperopt --hyperopt-loss SortinoHyperOptLoss --strategy Rainbow -c configusdt.json --eps
+# $ freqtrade hyperopt --hyperopt-loss SortinoHyperOptLossDaily --strategy Rainbow -c configusdt.json --eps
```
xmatthias, line 16: using `--eps` is missleading, as it'll end in results that cannot be recreated in live runs (which does not support position stacking).

i'd therefore not propose it here (you'll probably also need to reevaluate if the strategy is then still viable).

user_data/strategies/Rainbow.py
```diff
@@ -0,0 +1,216 @@
...
+
+        if conditions:
+            dataframe.loc[
+                reduce(lambda x, y: x | y, conditions), 'sell']=1
+
+        return dataframe
+
+    # you can add more conditions here with new tags and than add
+    # them to comperator hyperoptable variable.
+    def ConditionGenerator(self, comperator, df, A, B):
```
xmatthias, line 209: As mentioned before - please familiarize yourself with python naming conventions.

The naming (uppercase first letter) suggests a class-name (also the syntax-highlighting does) - but in reality it's a function.

user_data/strategies/Rainbow.py
```diff
@@ -8,12 +8,12 @@
...
-# $$$ freqtrade hyperopt --hyperopt-loss SharpeHyperOptLoss --strategy Rainbow -c configusdt.json --eps
-# $ freqtrade hyperopt --hyperopt-loss SharpeHyperOptLossDaily --strategy Rainbow -c configusdt.json --eps
-# $$ freqtrade hyperopt --hyperopt-loss SortinoHyperOptLoss --strategy Rainbow -c configusdt.json --eps
-# $ freqtrade hyperopt --hyperopt-loss SortinoHyperOptLossDaily --strategy Rainbow -c configusdt.json --eps
+# freqtrade hyperopt --hyperopt-loss ShortTradeDurHyperOptLoss --strategy Rainbow -c configusdt.json -e 10
+# $ freqtrade hyperopt --hyperopt-loss OnlyProfitHyperOptLoss --strategy Rainbow -c configusdt.json -e 10
+# $$$ freqtrade hyperopt --hyperopt-loss SharpeHyperOptLoss --strategy Rainbow -c configusdt.json -e 10
+# $ freqtrade hyperopt --hyperopt-loss SharpeHyperOptLossDaily --strategy Rainbow -c configusdt.json -e 10
+# $$ freqtrade hyperopt --hyperopt-loss SortinoHyperOptLoss --strategy Rainbow -c configusdt.json -e 10
+# $ freqtrade hyperopt --hyperopt-loss SortinoHyperOptLossDaily --strategy Rainbow -c configusdt.json -e 10
```
xmatthias, line 16: `-e 10` will not produce any meaningful results.
As mentioned for the Xth time to you: Please read the documentation!!!!

Please either:
* Remove command examples that don't make sense
* or make them meaningfull.

It's absolutely pointless to have samples for each possible hyperoptloss function - and your $$$ comments don't make sense (especially not with the parameters used) and show me that you are again not even trying / testing the code you're contributing.

user_data/strategies/Rainbow.py
```diff
@@ -0,0 +1,211 @@
...
+                    conditions.append(self.condition_generator(
+                        self.sell_ma_comperator.value, dataframe, dataframe['A'], dataframe['B']))
+
+        elif self.sell_ma_driver.value == 'shift':
+            rng = self.sell_ma_shift.range
+            for shift in rng:
+                dataframe['A'] = dataframe[self.sell_ma_time_range.value].rolling(
+                    self.sell_ma_rolling.value).mean().shift(shift)
+                dataframe['B'] = dataframe[self.sell_ma_time_range.value].rolling(
+                    self.sell_ma_rolling.value).mean().shift(shift-1)
```
xmatthias, line 192: The strategy is looking into the future, and therefore making one of the [common mistakes we have in our documentation](https://www.freqtrade.io/en/stable/strategy-customization/#common-mistakes-when-developing-strategies)

user_data/strategies/Rainbow.py
```diff
@@ -0,0 +1,211 @@
...
+    def populate_buy_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+        conditions = []
+
+        if self.buy_ma_driver.value == 'time_range':
+            rng = self.time_ranges
+            for i, tr in enumerate(rng):
+                if i-1 > 0:
+                    dataframe['A'] = dataframe.iloc[:, i].rolling(
+                        self.buy_ma_rolling.value).mean().shift(self.buy_ma_shift.value)
+                    dataframe['B'] = dataframe.iloc[:, i-1].rolling(
```
xmatthias, line 129: Here too (probably) - although i don't really understand what "i" can become.

mablue: if i-1 > 0 or if always i >= 1 
So there is no negative value execution.

xmatthias: this might be true for this instance, but not for the one pointed out above in the other comment.

mablue:
Multima is ok ! new Strategy Borned! :)

PR #225, opened 2021-09-23
Fix MultiMa to not give an error

Andy477:
Fixing buy_ma_count's Int Parameter value since they cause an error.
In more detail, the old version had buy_ma_count starting off at 2, then had a buy-ma plus 1. 
Then the smallest period(e.g dataframe["buy-ma-3"]), is 3.
But in the old code, there is dataframe[f"buy-ma-{i-1}"], if i is equal 3, then i - 1 is equal to 2. 
So forth causing an exception, as there is no dataframe["buy-ma-2"] as the smallest period is dataframe["buy-ma-3"]. 
The same thing happens if buy_ma_count's low value is equal to 1.
The same thing happens on the sell-side.
The only way to fix it is to have buy/sell_ma_count's low value set to 0.
There are also some formats(from the black formatter(https://github.com/psf/black)).

### DIFF
--- a/user_data/strategies/MultiMa.py
+++ b/user_data/strategies/MultiMa.py
@@ -7,6 +7,7 @@
 from freqtrade.strategy.hyper import IntParameter
 from freqtrade.strategy.interface import IStrategy
 from pandas import DataFrame
+
 # --------------------------------
 
 # Add your lib to import here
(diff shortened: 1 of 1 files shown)


xmatthias (approved):
I guess it'll be ok for now - but i think #213 will remove this strategy completely (it's author's decision after all).

PR #226, opened 2021-09-27
Try to reduce dataframe fragmented

bouroo:
use pd.concat(axis=1) instead of calling frame.insert many times

### DIFF
--- a/user_data/strategies/Supertrend.py
+++ b/user_data/strategies/Supertrend.py
@@ -16,6 +16,7 @@ import logging
 from numpy.lib import math
 from freqtrade.strategy.interface import IStrategy
 from freqtrade.strategy.hyper import IntParameter
+import pandas as pd
 from pandas import DataFrame
 import talib.abstract as ta
 import numpy as np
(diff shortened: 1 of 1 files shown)


user_data/strategies/Supertrend.py
```diff
@@ -80,39 +81,77 @@ class Supertrend(IStrategy):
...
+                    self.supertrend(df, multiplier, period)['STX'], columns=[f'supertrend_1_sell_{multiplier}_{period}']
+                ) for period in self.sell_p1.range
+                ], ignore_index=True)
+        pd.concat([dataframe, supertrend_1_sell], axis=1)
+        del supertrend_1_sell
 
         for multiplier in self.sell_m2.range:
-            for period in self.sell_p2.range:
-                dataframe[f'supertrend_2_sell_{multiplier}_{period}'] = self.supertrend(dataframe, multiplier, period)['STX']
+            supertrend_2_sell = pd.concat([
```
xmatthias, line 125: This is a very bad way of doing things - while it removes the performancewarning - performance will actually be worse, as you're now concating the dataframe X (range) amount of times.

The PerformanceWarning you're seeing is also fixed in stable - so this will no longer be necessary to get rid of that output.

Alternatively, look at [this documentation section](https://www.freqtrade.io/en/latest/strategy-advanced/#performance-warning) on how to do this properly.

xmatthias (changes requested):

PR #229, opened 2021-11-04
Fix typo in last fix

Daenks:
## Summary

prevents: freqtrade.resolvers.iresolver - WARNING - Could not import /home/$USERNAME/freqtrade/user_data/strategies/MultiMa.py due to 'invalid syntax (MultiMa.py, line 28)'


xmatthias (approved):
Great catch 👍

PR #230, opened 2021-11-05
MultiMa V2(Errors fixed+Shift removed+offline indicator population)

mablue:
## Summary
MultiMa V2(Errors fixed+Shift removed+offline indicator population)
Solve the issue: #___

## Quick strategy idea

- shifting droped
- indicators population moved inside the indicator populator.
- New algorithm with minimum coding :)

### DIFF
--- a/user_data/strategies/MultiMa.py
+++ b/user_data/strategies/MultiMa.py
@@ -1,8 +1,7 @@
-# MultiMa Strategy
+# MultiMa Strategy V2
 # Author: @Mablue (Masoud Azizi)
 # github: https://github.com/mablue/
-# (First Hyperopt it.A hyperopt file is available)
-#
+
 # --- Do not remove these libs ---
 from freqtrade.strategy.hyper import IntParameter
 from freqtrade.strategy.interface import IStrategy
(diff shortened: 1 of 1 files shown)


xmatthias (commented):
to be honest, i don't think updates like this should be done this way (with the same strategy/name).

as a user, i'd expect an update to be minor - fix some things - but continue to work mostly the same.

This is however basically a completely different strategy - especially since you changed the comparison for selling from and to or, and changed the MA type from SMA to TEMA, which has completely different characteristics.

The update itself seems okish - but i'd name/move it to "MultiMa_v2" - so it's clear that this is a different approach - and results from the previous strategy will not carry over.

mablue:
> to be honest, i don't think updates like this should be done this way (with the same strategy/name).
> 
> as a user, i'd expect an update to be minor - fix some things - but continue to work mostly the same.
> 
> This is however basically a completely different strategy - especially since you changed the comparison for selling from and to or, and changed the MA type from SMA to TEMA, which has completely different characteristics.
> 
> The update itself seems okish - but i'd name/move it to "MultiMa_v2" - so it's clear that this is a different approach - and results from the previous strategy will not carry over.

The biggest reason to this update is the errors happen on deferent run modes(hyperopt live backtesting)
Please check multima in defferent run modes. I make a mistake about using .range instade of range(...value) 

And also must of generated keys are overwriten cuz of mistake on naming
Cuz of this two big problem I commited this one .
But If you can fix this two problem with minimum changes on the orginal? Or I think multima is work good but have a useless forloop that just overwrites the indicators cuz of using equal key names  in some loops.

user_data/strategies/MultiMa.py
```diff
@@ -17,72 +16,84 @@
...
+    buy_ma_gap = IntParameter(1, gap_max, default=7, space="buy")
 
-        return dataframe
+    sell_ma_count = IntParameter(1, count_max, default=7, space="sell")
+    sell_ma_gap = IntParameter(1, gap_max, default=94, space="sell")
 
-    def populate_buy_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+    def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+        for count in range(self.count_max):
+            for gap in range(self.gap_max):
```
xmatthias, line 64: Please don't use `range(self.count_max)` - but use `self.count_max.range`.

That way, no unnecessary columns are added for backtesting/live runs - and are only populated during hyperopt.
You'll obviously need to make sure to properly address these down below - but with this - i'll not be accepting this, as it's not taking advantage of the optimizations freqtrade gives.

xmatthias (changes requested):

user_data/strategies/MultiMa.py
mablue: I know what you mean but I need full of range(0, X) on my decisions. The hyperopt just give us X to findout how much overlap conditions will appended to our strategy!
You know what I mean? I mean in this strategy we need exactly 5 conditions if our count_max is 5.
Also I think on this code I missed ".value" too

mablue: Yes I missed. I will add .value too. To get an static range from 0 to hyperopted number and use it every where.dose not matter where we use it in dry or live or backtest or hyperopt we need full range loops

xmatthias: ah yeah - missed that part ...

mablue: nooo its not missed part I got mistake!! it ok cuz `self.count_max` and `self.gap_max` are not hyperoptable params! :laughing: 
and also we need all range :) and just one thing we need changing name to V2 ! but MultiMa itself have some problems!

xmatthias (approved):

PR #231, opened 2021-11-05
Ichess Strategy added

mablue:
forex-comparic-photo-392-of-564
 Ichess Strategy  ichimoku_score
 Associate various ichimoku signals with a score. 
 For example, bullish signal => positive score, 
 and bearish signal => negative score. If the total score is above 0, it may indicate a bullish trend . 
 Otherwise, if it is below 0, it may indicate a bearish trend . 
 We used two smoothed moving averages to find the trend.
 More info:
 https://github.com/freqtrade/freqtrade-strategies/issues/97
 https://www.tradingview.com/script/P1bybHZA-Ichimoku-Cloud-Signal-Score-v2-0-0/

### DIFF
--- a/user_data/strategies/Ichess.py
+++ b/user_data/strategies/Ichess.py
@@ -1,8 +1,8 @@
 # Ichess Strategy  ichimoku_score
-# Associate various ichimoku signals with a score. 
-# For example, bullish signal => positive score, 
-# and bearish signal => negative score. If the total score is above 0, it may indicate a bullish trend . 
-# Otherwise, if it is below 0, it may indicate a bearish trend . 
+# Associate various ichimoku signals with a score.
+# For example, bullish signal => positive score,
+# and bearish signal => negative score. If the total score is above 0, it may indicate a bullish trend .
+# Otherwise, if it is below 0, it may indicate a bearish trend .
 # We used two smoothed moving averages to find the trend.
 # More info:
 # https://github.com/freqtrade/freqtrade-strategies/issues/97
@@ -11,9 +11,10 @@
 # github: https://github.com/mablue/
 
 # --- Do not remove these libs ---
-from freqtrade.strategy.hyper import IntParameter
+from freqtrade.strategy import IntParameter
 from freqtrade.strategy.interface import IStrategy
 from pandas import DataFrame
+import numpy as np
 # --------------------------------
 
 # Add your lib to import here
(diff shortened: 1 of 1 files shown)


user_data/strategies/Ichess.py
```diff
@@ -0,0 +1,260 @@
...
+        df['chikouPlacementScore'] = df.apply(calcChikouPlacement, axis=1)
+
+        # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # //
+        dataframe['Ichimoku_Score'] = (
+            df['tkCrossScore'] +
+            df['pkCrossScore'] +
+            df['kumoBreakoutScore'] +
+            df['senkouCrossScore'] +
+            df['chikouCrossScore']
+        ).cumsum()
```
xmatthias, line 242: this will not work identically during backtesting and live runs.
backtesting has all data available - so you'll calculate this on a range of potentially 100ds of days (the end value can be pretty high).
You should instead use `.rolling().cumsum()` (didn't check for other such problems in this strategy ...).

user_data/strategies/Ichess.py
```diff
@@ -0,0 +1,260 @@
...
+                if (intersect > x['senkou_span_a']) and (intersect > x['senkou_span_b']):
+                    return -0.5
+                elif(intersect < x['senkou_span_a']) and (intersect < x['senkou_span_b']):
+                    return -2
+                else:
+                    return -1
+            else:
+                return 0
+
+        df['tkCrossScore'] = df.apply(calcTkCross, axis=1)
```
xmatthias, line 129: This is a very slow way of doing things.
i don't see any reason not to do this in a vectorized manner - which will be a lot quicker.

user_data/strategies/Ichess.py
mablue: thanks but I got two errors:
`TypeError: rolling() missing 1 required positional argument: 'window'`
which window is best default for rolling?
after add a number `.rolling(9).cumsum()`
`AttributeError: 'Rolling' object has no attribute 'cumsum'`
Whats the best change for fix?

user_data/strategies/Ichess.py
mablue: oh than what can we do?!! you said befour i can use dt.hour! can I use some thing like dt.hour instade of that, here?

xmatthias: please do your own research - pandas is a wide enough topic so you can find probably 100ds of resources there - i'd suggest to properly learn pandas first.

xmatthias (changes requested):
`np.vectorize` is NOT what i meant, and will have no performance benefit over using `apply()`. it's the equivalent to `pd.apply` - which both execute the function n times (n = number of rows).


A vectorized approach is for example the following: `dataframe['whatever'] = dataframe['low'] / dataframe['high']` - which is operating on series, not on single values.

mablue:
> `np.vectorize` is NOT what i meant, and will have no performance benefit over using `apply()`. it's the equivalent to `pd.apply` - which both execute the function n times (n = number of rows).
> 
> A vectorized approach is for example the following: `dataframe['whatever'] = dataframe['low'] / dataframe['high']` - which is operating on series, not on single values.

Okay thanks👌👌👌 I think I will work more on it than ;)

mablue:
I reached good things about this strategy!
Now this strategy splited to two versions and I just want to keep one of them! The one that is insane! But the normal one is good and simple!-_-

froggleston:
I see no updates to this PR that address the review so would recommend this is closed and reopened if any further work has been done.

PR #232, opened 2021-11-05
Khayyam strategy added

mablue:
#### Khayyam Strategy
 In this strategy we try to find the best hours to buy and sell in a day.(in hourly timeframe)
 Cuz of that you should just use 1h timeframe on this strategy.
 Name of this strategy come from the Omar khayyam who was a Persian polymath, 
 mathematician, astronomer, philosopher, and poet! 
 https://en.wikipedia.org/wiki/Omar_Khayyam
 freqtrade hyperopt --hyperopt-loss SharpeHyperOptLoss --strategy Khayyam

<p dir="rtl" lang="fa">
اسرار ازل را نه تو دانی و نه من<br/>
وین حرف معما نه تو خوانی و نه من<br/>
هست از پس پرده گفت و گوی من و تو<br/>
چون پرده برافتد،نه تو مانی و نه من<br/>
</p>

Neither you nor I know the secrets of eternity<br/>
Neither you nor I can read this riddle<br/>
From behind the curtain<br/>
There is a conversation between you and me <br/>
When the curtain falls, Neither you nor I, will exist!<br/>

### DIFF
--- a/user_data/strategies/Khayyam.py
+++ b/user_data/strategies/Khayyam.py
@@ -1,60 +1,108 @@
 # Khayyam Strategy
 # In this strategy we try to find the best hours to buy and sell in a day.(in hourly timeframe)
 # Cuz of that you should just use 1h timeframe on this strategy.
-# Name of this strategy come from the Omar khayyam who was a Persian polymath, 
-# mathematician, astronomer, philosopher, and poet! 
+# Name of this strategy come from the Omar khayyam who was a Persian polymath,
+# mathematician, astronomer, philosopher, and poet!
 # https://en.wikipedia.org/wiki/Omar_Khayyam
 # Author: @Mablue (Masoud Azizi)
 # github: https://github.com/mablue/
-# freqtrade hyperopt --hyperopt-loss SharpeHyperOptLoss --strategy Khayyam
+# * RUN LIVE AFTER HYPEROPT!
+# freqtrade hyperopt --hyperopt-loss SharpeHyperOptLoss --strategy Khayyam -e 200
 
-from freqtrade.strategy.hyper import IntParameter
+
+from freqtrade.strategy import IntParameter
 from freqtrade.strategy import IStrategy
 from pandas import DataFrame
 # --------------------------------
 # Add your lib to import here
+# No need to These imports. just for who want to add more conditions:
+# import talib.abstract as ta
+# import freqtrade.vendor.qtpylib.indicators as qtpylib
 
 
 class Khayyam(IStrategy):
+    # SHIB/USDT, 1000$x1:100days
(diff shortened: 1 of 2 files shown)


user_data/strategies/Khayyam.py
```diff
@@ -0,0 +1,60 @@
...
+    sell_hour_max = IntParameter(0, 24, default=6, space='sell')
+    sell_hour_min = IntParameter(0, 24, default=15, space='sell')
+
+    def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+        return dataframe
+
+    def populate_buy_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+        dataframe.loc[
+            (
+                dataframe['date'].apply(lambda x: x.hour).between(
```
xmatthias, line 45: i think you should avoid using `.apply` when possible.
It's the equivalent of a loop through the whole dataframe, which will be implicitly slow.

Best extract the value you need to a different column, and compare to that directly
(in demonstrative, non-working code):
``` python
dataframe['hour'] = dataframe['date'].dt.hour
# ...
min_val < dataframe['hour'] < max_val
```

user_data/strategies/Khayyam.py
```diff
@@ -1,60 +1,108 @@
...
+        dataframe['hour'] = dataframe['date'].dt.hour
         return dataframe
 
     def populate_buy_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+        min, max = self.buy_hour_min.value, self.buy_hour_max.value
         dataframe.loc[
             (
-                dataframe['date'].apply(lambda x: x.hour).between(
-                    self.buy_hour_min.value, self.buy_hour_max.value)
+                (dataframe['hour'].between(min, max))
```
xmatthias, line 95: Are you sure this works? 
i'd not bet on how pandas behaves when using `between(1, 0)` ... which are the default parameters you define for this strategy.

mablue: yes I know cuz I changed this defaults manually.
I want to everybuddy hyperopt befour run at live.
you know, Cuz Trading  with just watching to the clock is a vary dangerous approach. after running just one hyper-opt we can prove some of protections will be active on trading like stop loss and roi and etc...I disabled manually, cuz It can take users to the high risks

xmatthias (approved):
I think strategy naming should make it clear what a strategy does.
the name here suggests some secret combination of indicators that desires a name - while in reality, it's just time-based trading.

Because of this, i see the naming as inapprorpiate (not clear to what it does), and have therefore changed it to `HourBasedStrategy`.

mablue:
Yes thats the way thanks for accept I will try to dont break this rules again ;)

PR #249, opened 2022-05-20
Add multi_tf.py

stash86:

## Summary

Add example strat that uses different timeframes

Solve the issue: #___

## Quick strategy idea

- <change log 1>
- <change log 2>


xmatthias (approved):
Nice :+1:

PR #262, opened 2022-08-08
updated buy and sell parameters

TreborNamor:

## Summary

The goal is to inspire people with variety of strategies.
I added ema parameters to see if it had a better influence on trading strategy.

Solve the issue: #___

## Quick strategy idea

- <change log 1>
- <change log 2>

### DIFF
--- a/user_data/strategies/Bandtastic.py
+++ b/user_data/strategies/Bandtastic.py
@@ -1,0 +1,159 @@
+import talib.abstract as ta
+import numpy as np  # noqa
+import pandas as pd
+from functools import reduce
+from pandas import DataFrame
+import freqtrade.vendor.qtpylib.indicators as qtpylib
+from freqtrade.strategy.interface import IStrategy
+from freqtrade.strategy import CategoricalParameter, DecimalParameter, IntParameter, RealParameter
+
+__author__ = "Robert Roman"
+__copyright__ = "Free For Use"
+__license__ = "MIT"
+__version__ = "1.0"
+__maintainer__ = "Robert Roman"
+__email__ = "robertroman7@gmail.com"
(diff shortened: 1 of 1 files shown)


user_data/strategies/Bandtastic.py
```diff
@@ -0,0 +1,159 @@
...
+        conditions = []
+
+        # GUARDS
+        if self.buy_rsi_enabled.value:
+            conditions.append(dataframe['rsi'] < self.buy_rsi.value)
+        if self.buy_mfi_enabled.value:
+            conditions.append(dataframe['mfi'] < self.buy_mfi.value)
+        if self.buy_ema_enabled.value:
+            try:
+                conditions.append(ta.EMA(dataframe, timeperiod=int(self.buy_fastema.value)) > ta.EMA(dataframe, timeperiod=int(self.buy_slowema.value)))
```
xmatthias, line 103: I know it works this way - but i think sample strategies should kind of follow the approach the [documentation](https://www.freqtrade.io/en/latest/hyperopt/#optimizing-an-indicator-parameter) explains.

If you think it's problematic to use the `.range` approach (that's usually a lot faster when hyperopting - at least if done correctly), at least assign the EMA values to columns - so they can used in plotting. 

Currently all EMA's are hidden, non-visible features that'll be causing buys/sells (or non-buy/sells) - which is obviously problematic for users who don't know about this, and will have problems understanding this segemnt.

xmatthias (changes requested):

froggleston:
Should this be closed? I don't see any activity or really any benefit to adding this code to this example.

xmatthias (approved):
I've made the requested changes.

if memory in hyperopt is a problem, we can use `--analyze-per-epoch` to avoid calculating all EMA's at once - though i don't see that as a problem, as it'll be a max. of 365 columns based on current settings.

Issue #284, opened 2022-12-30
Lookahead for Strategy002 and 004


froggleston:
## Describe your environment

  * Strategy002: https://github.com/freqtrade/freqtrade-strategies/blob/main/user_data/strategies/Strategy002.py#L103
  * Strategy004: https://github.com/freqtrade/freqtrade-strategies/blob/main/user_data/strategies/Strategy004.py#L107
  
## Describe the problem:

I'm concerned about the talib pattern functions, e.g. CDLHAMMER used in Strategy002. It seems that the talib function has an empty lookback value, so I presume this uses the whole dataframe in backtesting, so has lookahead?

https://github.com/mrjbq7/ta-lib/blob/a42ba010bb6f983883dd2d07668f82bc82b4c236/talib/_func.pxi#L1466

```
    lookback = begidx + lib.TA_CDLHAMMER_Lookback( )
    outinteger = make_int_array(length, lookback)
```

If we compare this to something like LINEARREG which has a defined timeperiod:

```
    lookback = begidx + lib.TA_LINEARREG_Lookback( timeperiod )
    outreal = make_double_array(length, lookback)
```

In a more certain case, Strategy 004 in the main example repo has the following line:

`dataframe['mean-volume'] = dataframe['volume'].mean()`

This will have lookahead in backtesting. Either this strategy should be updated to a rolling implementation or a comment explicitly stating that this strategy will have differing backtest vs dry/live results.

xmatthias:
with Strategy004 it's clear - we'll be fixing this.

for the "hammer" issue - i'm not sure.
I've never used pattern functions before - so i'm not really certain on how they work.

The same goes with the function you point out - i struggled to find the definition for it - so it **may** just use "everything up to now" ... which would kinda be fine (it's not lookahead per se) - though it'd still be a form of problem, as it assumes an (indefinitely) expanding window.

froggleston:
I'm happy to raise a PR for 004.

Yeah, with the pattern functions, it's unclear as to what an empty lookback function actually does in talib...

xmatthias:
well technically - i think the lookback is pretty much given by the indicator itself.

a hammer is just 1 candle - in a pattern, something between 3 and 5 (depending if you wait for confirmation) - so even if it were to use expanding windows - a candle 200 lines from the pattern wouldn't count.

froggleston:
Yeah I started a thread in the main discord about it for the candle pattern stuff, but mainly focused around these two SO posts:

[...]

PR #285, opened 2022-12-31
Remove volume mean lookahead

froggleston:
Add a rolling example to fix the backtesting lookahead, and add a note that the original code would be fine for dry/live use.

Solve the issue: https://github.com/freqtrade/freqtrade-strategies/issues/284


xmatthias (approved):
Great! thanks! :+1:

PR #286, opened 2023-01-07
add TrendFollowingStrategy

PedroPovoleri:

## Summary

Create a new strategy for future markets TrendFollowingStrategy

## Quick strategy idea

This commit adds the TrendFollowingOBVStrategy, which is a trend-following strategy that uses the EMA and OBV indicators to generate buy and sell signals. The strategy implements the IStrategy interface and includes the populate_indicators, populate_entry_trend, and populate_exit_trend functions. The strategy has been tested and is ready for use.


user_data/strategies/futures/TrendFollowingStrategy.py
```diff
@@ -0,0 +1,52 @@
+from functools import reduce
+from pandas import DataFrame
+from freqtrade.strategy import IStrategy
+
+import talib.abstract as ta
+
+from freqtrade.strategy.interface import IStrategy
+
+class TrendFollowingStrategy(IStrategy):
+    def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
```
xmatthias, line 10: we would clearly want "some" of the strategy attributes defined here.


they're not mandatory to have in the strategy (they can be loaded from the configuration) - but most strategies have it configured - first off, to demonstrate "this should run on timeframe X, with stoploss y" - and then because i see strategies as something that can change more often than configurations - so all "strategy related" things are in one place.
without these, it's just a strategy without ANY guidance.

i think what we'd want is the 3 main ones
* `timeframe`
* `minimal_roi`
* `stoploss`
* (`trailing_stop`) (if you think that's used / necessary).

PedroPovoleri: Sure I will change it for that points, thanks for the review [xmatthias](https://github.com/xmatthias)

xmatthias (approved):

PR #287, opened 2023-01-08
Add volatility stratgegy

### LINKED ISSUE #283: Volatility Strategy
## Step 1: What indicators are required?

Here is a link to the strategy which uses modified ATR with a pinescript.
https://www.tradingview.com/script/3hhs0XbR/

## Step 2: Explain the Buy Strategy

Code provided above

## Step 1: Explain the Sell Strategy

Code provided above

## Source

What come from this strategy? Cite your source:

https://www.tradingview.com/script/3hhs0XbR/

Other Strategy modifications:

Strategy works best on a 3h timeframe and I am not sure how to work around the timeframes provided by exchange.

They workflow of the strategy should be like this. 

1- Long/short entry signal found for a pair, bot should place an order of 50% of the stake amount.
2- If another entry signal is found on the same direction, it should place the other 50%.
3- If a long/short exit entry is found, bot will exit full position and simultaneously place an order on the opposite direction. at least this is how it works on tradingview. 
- eg. bot found an entry long signal for BTCUSDT at 11000 and stake amount is $100, bot will place an order for $50. Bot found another entry long signal at 12000 then bot will place another $50 in the trade and stop trading in this direction.
- Bot found an exit signal at 12500, bot will exit full position at 12500 and then enter a short position at 12500 for 50% of the stake amount.

Thanks

xmatthias:
## Summary
Volatility based strategy - as defined in the request / tradingview source.

includes 
* scale into position
* long/short

closes #283

## Quick strategy idea

Replication of https://www.tradingview.com/script/3hhs0XbR/


user_data/strategies/futures/VolatilityStrategy.py
```diff
@@ -0,0 +1,137 @@
...
+        dataframe.loc[
+            dataframe['enter_short'] == 1,
+            'exit_long'] = 1
+        return dataframe
+
+    def custom_stake_amount(self, pair: str, current_time: datetime, current_rate: float,
+                            proposed_stake: float, min_stake: Optional[float], max_stake: float,
+                            leverage: float, entry_tag: Optional[str], side: str,
+                            **kwargs) -> float:
+        # 50% stake amount on initial entry
```
FahadAziz, line 113: There is an issue here as the bot enters two trades on the same candle

user_data/strategies/futures/VolatilityStrategy.py
```diff
@@ -0,0 +1,137 @@
...
+    position_adjustment_enable = True
+
+    def adjust_trade_position(self, trade: Trade, current_time: datetime,
+                              current_rate: float, current_profit: float,
+                              min_stake: Optional[float], max_stake: float,
+                              current_entry_rate: float, current_exit_rate: float,
+                              current_entry_profit: float, current_exit_profit: float,
+                              **kwargs) -> Optional[float]:
+        dataframe, _ = self.dp.get_analyzed_dataframe(trade.pair, self.timeframe)
+        # Only buy when not actively falling price.
```
FahadAziz, line 125: The second entry should be entered if there is a second signal regardless of the price direction compared to the last signal

user_data/strategies/futures/VolatilityStrategy.py
```diff
@@ -0,0 +1,137 @@
...
+            (dataframe['close_change'] * -1 > dataframe['atr'].shift(1)),
+            'enter_short'] = 1
+
+        return dataframe
+
+    def populate_exit_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+        """
+        use sell/buy signals as long/short indicators
+        """
+        dataframe.loc[
```
FahadAziz, line 101: would it make sense to reverse the entry signals to work around exiting and entering on the same candle?

user_data/strategies/futures/VolatilityStrategy.py
xmatthias: can you explain what you mean with this?

freqtrade won't open 2 trades on the same pair on the same candle. That'd be considered a bug - and nothing in this strategy is novell or trying to circumvent anything in that regards.

user_data/strategies/futures/VolatilityStrategy.py
xmatthias: the comment is wrong 😆 
the real comparison is to only act on the "first" signal (when it flicks from "nothing" to "buy"/"sell").

that's a "workaround" as we use 3h candles - so the signal is in reality active throughout all 3 hours - but we don't want to buy 2 hours into a 3 hour candle.  That signal is clearly outdated at that point.

user_data/strategies/futures/VolatilityStrategy.py
xmatthias: Freqtrade does not support reversing a trade at this point.

That's nothing this strategy will change or can do to work around that.

user_data/strategies/futures/VolatilityStrategy.py
FahadAziz: Screenshot 2023-01-10 at 3 22 51 PM

This is the first trade the bot created. it entered two orders on the same price

user_data/strategies/futures/VolatilityStrategy.py
FahadAziz: I tried it and it gives different results than the original code. I will dig deeper in the numbers and see what is happening

user_data/strategies/futures/VolatilityStrategy.py
FahadAziz: Screenshot 2023-01-10 at 3 39 07 PM
Screenshot 2023-01-10 at 3 38 49 PM


Here I am trying to replicate the backtesting results on tradingview (configured to 1 trade instead of two). Similar results

user_data/strategies/futures/VolatilityStrategy.py
xmatthias: well the original code assumes a positions  switch - which is not supported by freqtrade - so i consider out of scope here.

user_data/strategies/futures/VolatilityStrategy.py
xmatthias: interesting - well this is clearly not how i thought it 😆

xmatthias: should be gone now - at least mine doesn't seem to do this anymore.

FahadAziz: I don't know what happened here, but here are my observations. 

1- bot always enters one trade with 50% stake amount. I think this is because it cannot find a successive entry since tradingview is counting the reverse trade.
2- Here are exact snapshots comparison between tradingview and freqtrade of trades that should have happened but freqtrade is showing exit existing trades that do not exist while it should enter a long trade.

Screenshot 2023-01-12 at 11 47 42 AM
Screenshot 2023-01-12 at 11 42 53 AM

These are my current observations

user_data/strategies/futures/VolatilityStrategy.py
FahadAziz: Is it because of the auto lock? can we unlock the pair after it exits a trade to open it for the reverse trade?

user_data/strategies/futures/VolatilityStrategy.py
FahadAziz: I made some changes and let it run on dry run and see how it goes to understand the behavior better.

user_data/strategies/futures/VolatilityStrategy.py
xmatthias: auto-lock is implicit in backtesting.

user_data/strategies/futures/VolatilityStrategy.py
xmatthias: Scaling works fine for me.

The most likely reason is that your backtest ran out of money.

image


You also should be careful as the UI will render symbols atop of each other.
every enter_long is at the same time an exit_short. 
If you disable exit_short in the legend, you'll observe the hidden enter_long signal.

PR #289, opened 2023-02-09
Set process candles to True

froggleston:
## Summary

There's no reason for `process_only_new_candles` to be False in this strategy example as there's no ticker/websocket data being used, just candles.


xmatthias (approved):
This is probably true for most strategies in this repository ... :laughing: 

:+1:

PR #290, opened 2023-02-12
Universal MACD Strategy Added!

mablue:
## Summary

I changed [MACD](https://www.tradingview.com/scripts/macd/) formula to divergence of (MA12/MA26 - 1).
And its make it more useful.
Cuz:
1) comparability with all other coins with different prices.
2) fix small numbers in low price coines like shiba
3) making a almost limited-normalized indicator like RSI to use it for optimization and ML/AI projects as a variable

Most important thing about this indicator is that its Universal
Now you can compare the UMACD of Shiba with btc without any problem in matamatics space.No need to use virtuality and its important in Optimization problems that we rediuse the problem from a picture to a number(A plot to a list of numbers)
If we don't care about exagrated pumps and dumps, we can say to it Normalized-MACD too. cuz in normal situations its MAX ≈ 0.1 and MIN ≈ -0.1 in daily tf
https://www.tradingview.com/script/xNEWcB8s-Universal-Moving-Average-Convergence-Divergence/

## Quick strategy idea
the idea is that we try to find things like bricks when we want to make a house in the nature.
cuz the shape is important to keep things Over each other and construction.
the optimization algos are like that we try to solve a problem on nature and construct a formula to use it.
the bricks of the optimization algos are variables like RSI that are limited to 0,100 and/or any other variables and indicators that not infinite.
we try on UMACD to make an almost infinite indicator to use it as a limited variable. and solve the secret of getting rich!‌ :rofl:


user_data/strategies/UniversalMACD.py
```diff
@@ -0,0 +1,122 @@
...
+    buy_umacd_min = DecimalParameter(-0.05, 0.05, decimals=5, default=-0.01416, space="buy")
+    sell_umacd_max = DecimalParameter(-0.05, 0.05, decimals=5, default=-0.02323, space="sell")
+    sell_umacd_min = DecimalParameter(-0.05, 0.05, decimals=5, default=-0.00707, space="sell")
+
+    def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+        dataframe['ma12'] = ta.EMA(dataframe, timeperiod=12)
+        dataframe['ma26'] = ta.EMA(dataframe, timeperiod=26)
+        dataframe['umacd'] = (dataframe['ma12'] / dataframe['ma26']) - 1
+
+        print(dataframe['umacd'].min(), dataframe['umacd'].max())
```
xmatthias, line 102: pointless and way to verbose. Please remove.

user_data/strategies/UniversalMACD.py
```diff
@@ -0,0 +1,122 @@
...
+    # Strategy parameters
+    buy_umacd_max = DecimalParameter(-0.05, 0.05, decimals=5, default=-0.01176, space="buy")
+    buy_umacd_min = DecimalParameter(-0.05, 0.05, decimals=5, default=-0.01416, space="buy")
+    sell_umacd_max = DecimalParameter(-0.05, 0.05, decimals=5, default=-0.02323, space="sell")
+    sell_umacd_min = DecimalParameter(-0.05, 0.05, decimals=5, default=-0.00707, space="sell")
+
+    def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+        dataframe['ma12'] = ta.EMA(dataframe, timeperiod=12)
+        dataframe['ma26'] = ta.EMA(dataframe, timeperiod=26)
+        dataframe['umacd'] = (dataframe['ma12'] / dataframe['ma26']) - 1
```
xmatthias, line 100: take your pick - the description says it's `(ma26/ma12) -1` ... ? 

so reading the description, this is a bug? (or is the description wrong??)

user_data/strategies/UniversalMACD.py
mablue: just for show user the min and max of indicator in different coins to set inside hyperoptable variables. cuz in different timeframes should change the min and max in hyperoptable variables.
It can usable for users

user_data/strategies/UniversalMACD.py
mablue: ops I think discription is wrong I just changed macd formula : `ma12-ma26` to `(ma12/ma26)-1`

user_data/strategies/UniversalMACD.py
```diff
@@ -0,0 +1,125 @@
...
+    buy_params = {
+        "buy_umacd_max": -0.01176,
+        "buy_umacd_min": -0.01416,
+    }
+
+    # Sell hyperspace params:
+    sell_params = {
+        "sell_umacd_max": -0.02323,
+        "sell_umacd_min": -0.00707,
+    }
```
xmatthias, line 69: this part is redundant - and not really necessary.

It's the same numbers than the default - and freqtrade will override these values with what it loads from a param.json file anyway. best remove this whole section.

mablue: yes It's true I removed this dict params

xmatthias (approved):

PR #291, opened 2023-02-12
Tree Black Crows Strategy added!

mablue:
## Summary
3black crows is a candlistic pattetrn I based on pow and in this strategy I try to find optimom pow for this candlistic pattern to avoid big pumps and dumps.
## Quick strategy idea
I used the main idea of 3bc ( https://www.investopedia.com/terms/t/three_black_crows.asp )
but I used two decimal optimizable number to find best power to these 3 candles

### DIFF
--- a/user_data/strategies/PowerTower.py
+++ b/user_data/strategies/PowerTower.py
@@ -18,9 +18,10 @@ import pandas_ta as pta
 from technical import qtpylib
 
 
-class ThreeBlackCrows(IStrategy):
+class PowerTower(IStrategy):
     # By: Masoud Azizi (@mablue)
-    # Investopedia Link: https://www.investopedia.com/terms/t/three_black_crows.asp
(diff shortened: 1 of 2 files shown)


xmatthias (changes requested):
I'd think that the name of the strategy is highly misleading.
This is not even attempting to identify three black crows - it's simply a comparison of 3 (semi-random) candles.

It may accidentally detect a 3 black crows pattern (but on the sell side) - but that's far from the only pattern with the logic provided (letalone the fact that 3 black crows are 3 red candles - your entry signal checks for 3 green candles).

If the strategy yields good results for you in dry-run - i'm happy to accept this with a different name - but surely not with a "pattern name" - which is not used within the strategy.

mablue:
PowerTower is present! 💝 Thanks For your advice.

xmatthias (approved):

PR #303, opened 2023-07-15
Update example strategies to use process_only_new_candles=True

froggleston:
## Summary

A lot of the example strategies have been set to use `process_only_new_candles=False`, which will not provide any functionality as they only use OHLCV data. This PR sets all cases of `process_only_new_candles` to `True`


xmatthias (approved):
Great spot! :+1:

PR #309, opened 2024-04-21
Update Bandtastic.py

contributorINF113:
Updated populate_entry_trend and populate_exit_trend to follow new practices

## Summary

Solve the issue: #___

## Quick strategy idea

- <change log 1>
- <change log 2>

### DIFF
--- a/user_data/strategies/Bandtastic.py
+++ b/user_data/strategies/Bandtastic.py
@@ -98,7 +98,7 @@ class Bandtastic(IStrategy):
 
         return dataframe
 
-    def populate_buy_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+    def populate_entry_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
         conditions = []
 
         # GUARDS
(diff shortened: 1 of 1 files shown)


user_data/strategies/Bandtastic.py
```diff
@@ -129,7 +129,7 @@ def populate_buy_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
 
         return dataframe
 
-    def populate_sell_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+    def populate_exit_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
```
xmatthias, line 132: well if we assume to update this part - then we should follow through with all of the [strategy migration guide](https://www.freqtrade.io/en/stable/strategy_migration/) - which doesn't just include this change - but also others (for example updating the dataframe columns to enter_long instead of buy - just to name one without checking for other points).

xobazovv97-hash: https://github.com/freqtrade/freqtrade-strategies/pull/309/files

PR #330, opened 2026-01-25
feat: add LunarPhaseStrategy - lunar cycle and WuXing trading signals

sjhddh:
## Summary

Add a new strategy `LunarPhaseStrategy.py` that uses Chinese lunar calendar data to generate trading signals based on ancient metaphysics - a "cyber-mysticism" approach to trading.

## Features

| Component | Description |
|-----------|-------------|
| **Moon Phase** | Score based on lunar day (1-30). New Moon = bullish, Full Moon = neutral/bearish |
| **WuXing (五行)** | Five Elements theory. Wood/Fire = bullish for crypto, Water = bearish |
| **Heavenly Stems (天干)** | Additional timing signals based on the day's celestial stem |
| **Feng Shui Score** | Composite score (-10 to +10) combining all factors |

## Philosophy

> "Crypto markets are driven by human emotion. The moon controls human emotion. You do the math."

This is a HODL-style strategy that aligns entries with lunar accumulation phases (New Moon) and exits with distribution phases (Full Moon).

## Installation

```bash
pip install lunar_python
```

## Usage

```bash
freqtrade backtesting --strategy LunarPhaseStrategy
freqtrade trade --strategy LunarPhaseStrategy --dry-run
```

## Hyperoptable Parameters

- `entry_threshold`: Feng shui score threshold for entry (default: 5.0)
- `exit_threshold`: Feng shui score threshold for exit (default: -5.0)
- `moon_weight`: Weight for moon phase scoring (default: 1.0)
- `element_weight`: Weight for WuXing element scoring (default: 1.0)


limex:
Is this a niche belief or is trading based on moon phases commonly practiced in Asia?

sjhddh:
yes, You get it exactly. It is a thing. And people who believe in it will actually trade based on such timeline guidance, which we will call a “黄道吉日”; so, yes, if people believe in it, then you know there will be traders following such philosophies. In this sense, it is indeed a signal. An hidden alternative signal.

user_data/strategies/LunarPhaseStrategy.py
```diff
@@ -0,0 +1,538 @@
...
+# Install with: pip install lunar_python
+try:
+    from lunar_python import Solar, Lunar
+    LUNAR_AVAILABLE = True
+except ImportError:
+    LUNAR_AVAILABLE = False
+    logging.warning(
+        "lunar_python not installed. Please install with: pip install lunar_python\n"
+        "LunarPhaseStrategy will use fallback calculations."
+    )
```
xmatthias, line 70: Does this strategy also work without this package?
Are results identical? 
If they are - why would you then use / require it, if it isn't important??
If they are not - then this is even worse.


I think allowing both is a risk for users i'd rather not show as "sample" in this repository.

assume the following scenario:
* you backtest on your system, with lunar_python installed - and get good results
* you deploy your strategy to a server (maybe via docker) - but forget to install `lunar_python` - your results are completely different.

user_data/strategies/LunarPhaseStrategy.py
```diff
@@ -0,0 +1,538 @@
...
+        dataframe["wuxing"] = ""
+        dataframe["wuxing_english"] = ""
+        dataframe["tiangan"] = ""
+        dataframe["moon_score"] = 0.0
+        dataframe["wuxing_score"] = 0.0
+        dataframe["feng_shui_score"] = 0.0
+        dataframe["is_auspicious"] = False
+
+        # Calculate lunar data for each candle
+        for idx in dataframe.index:
```
xmatthias, line 401: i assume this will be pretty slow - especially in backtesting ... 
what do runtime metrics (like - how long does 1 pair with the default 1000 candle history) take to run this? 

That's what'll happen for every pair on every new candle - which is why i think that's important to at least note (docstring -> "strategy may be slow for backtesting" or similar) and possibly optimize.

xmatthias (approved):

xmatthias (changes requested):
Approved by mistake ... sorry for the confusion (extension going wild :laughing: )

i don't think that the "Optionality" of the dependency makes sense. 
Take your pick, either it's required - or it isn't.
Having 2 variants of a strategy within one - where the selection of which depends on arbitrary things (like in this case, did a depenency load) is not a stable way - and will lead to tons of confusion - up to the point of "why does backtstesting deviate from dry so much" - which is in reality caused by different environments with different dependencies.

xmatthias (commented):
I don't think

xmatthias (changes requested):
I don't think

xmatthias (commented):
I don't think

xmatthias (changes requested):
hello

xmatthias (approved):
asdfasdfasdf

xmatthias (changes requested):
asdfsdf

xmatthias (commented):
df

xmatthias (changes requested):
hell

xmatthias (commented):
dd

sjhddh:
Closing this PR to clean up backlog. Realized this doesn't meet the high-signal threshold for the project. Apologies for the noise!

PR #332, opened 2026-03-08
Add System R risk-managed momentum strategy

ashimnandi-trika:
## Summary

Adds a momentum strategy (`SystemRRiskManaged`) that validates every trade through [System R's](https://agents.systemr.ai) pre-trade risk gate before execution.

- Uses `confirm_trade_entry` callback to call System R's `pre_trade_gate` API
- EMA crossover + RSI for entry signals
- G-formula position sizing + Iron Fist risk validation in one API call
- Fail-open: if System R is unavailable, trades proceed normally (safe for backtesting)

## How it works

Before every trade entry, the strategy calls:
```
POST https://agents.systemr.ai/v1/compound/pre-trade-gate
```

If `gate_passed: true`, the trade executes. If blocked, the trade is skipped with a log message.

## Setup

```bash
pip install httpx
export SYSTEMR_API_KEY=your_key  # Free: 10,000 calls on registration
```

Register: https://agents.systemr.ai/v1/agents/register

Demo repo: https://github.com/System-R-AI/demo-trading-agent


stash86:
For any example strategy, we won't ask users to additionally install anything extra

ashimnandi-trika:
*Thanks for the review, Stefano. Totally understand the policy on keeping
strategies self-contained. Appreciate the time!*

xmatthias:
I think more important than the additional dependency (httpx is already installed implicitly) is the fact that it's calling a random, commercial service. 

This repository is here to serve as example freqtrade strategies for users to get started - not to promote some random AI service.

Happy to accept a strategy that's self-contained - but not one that's calling "random service XYZ".

PR #334, opened 2026-04-15
Add TrendRiderStrategy: V4 cascading early loss cut (66.7% WR, 1.42% DD on 30d Bybit)

darkvolg:
Hi maintainers — opening this PR to contribute a wide-stop trend-following strategy that I've been running publicly on a live dashboard.

## What it is

`TrendRiderStrategy.py` — a 1h-timeframe trend follower for Bybit USDT-perps across 15 large-cap pairs (BTC, ETH, SOL, BNB, XRP, DOGE, ADA, LINK, DOT, AVAX, ATOM, NEAR, OP, POL, SUI). Six-signal confidence score (EMA cross + RSI + ADX + volume z-score + Bollinger position + MACD histogram) with a tunable threshold gating entries.

Three enter_tags: `ema_crossover`, `macd_reversal`, `bb_bounce`. Standard Freqtrade ROI table + 6% stoploss + trailing stop with offset.

## What's interesting

The exit logic uses **cascading early loss cut** instead of the more common flat `time_exit_24h` rule. Two thresholds in `custom_exit`:

```python
if hours_open >= 4 and current_profit < -0.015:
    return "early_loss_cut_4h"
if hours_open >= 2 and current_profit < -0.025:
    return "early_loss_cut_2h"
```

The intuition: kill on **unrealized loss + time**, not time alone. A trade that's flat at hour 2 deserves more time. A trade that's down 2.5% at hour 2 is already telling you it was wrong.

## Backtest deltas (same 30-day window, same entries, only the exit rule changed)

| Metric | V3 (flat time_exit_24h) | **V4 (cascading)** |
|---|---|---|
| Profit factor | 1.03 | **1.41** |
| Max drawdown | −6.12% | **−1.42%** |
| Sharpe | 0.22 | **0.91** |
| Win rate | 66.7% | 66.7% (unchanged) |
| Avg loser | −$1.14 | **−$0.34** |

The strategy wasn't picking bad trades — it was holding bad trades wrong.

## What's NOT in this file

The production bot has a few private layers (Fear & Greed Index API, Bybit funding-rate signals, on-chain whale alerts, SQLite price-alert layer) which are stubbed out here with neutral defaults. The public version still trades cleanly, just without those external boosts.

## Live verification

The live bot runs this strategy in dry-run mode and exposes its SQLite publicly (read-only, every 60s):

- **Live dashboard**: https://trendrider.net/live
- **Source repo (MIT, with banner)**: https://github.com/darkvolg/trendrider-strategy
- **Full post-mortem with every trade**: https://trendrider.net/blog/freqtrade-bot-14-days-breakeven-v4-fix-2026

Anyone can clone the repo, drop in the same params, and reproduce the 30-day backtest.

## Caveats

- Tuned for current market regime (BTC $60-75k, RV ~3.5%). Will need re-hyperopt in different vol regimes.
- 30 days of backtest data — robust stats need 180+ days. Next re-hyperopt scheduled at 90 days.
- Dry-run only as of writing. Live $100 flip planned April 20 pending another 5 days of V4 validation.

## What I updated

- Added `user_data/strategies/TrendRiderStrategy.py` (745 lines, all docstrings preserved)
- Added one row to the README backtest table

Happy to discuss the cascade design or rework anything that doesn't fit the repo conventions. If you'd prefer a smaller / single-file version without the external-API stubs, let me know.

Thanks for Freqtrade — it's been an incredible learning environment.

— darkvolg


user_data/strategies/TrendRiderStrategy.py
```diff
@@ -0,0 +1,745 @@
+"""
+TrendRider Public v2.11.0 — Strat Ninja Edition
+
+Philosophy: Ride established trends with WIDE stoploss.
+Key insight: crypto swings 2-4% per hour. Stoploss must be >= 5-6%.
+
+Public version:
+- No external API calls (FNG, Bybit funding/OI)
+- No SQLite price alerts
+- No Cornix formatting
```
xmatthias: I don't think we need these 3/4 lines. 
It's a pointless comment which the code itself shows as well - and it implies that there's a "non-public" version.

While that in itself is fine (everyone has a right for their own private strategy) - the sample repository doesn't need to imply this.
Also the "No Cornix formatting" seems pointless here ...

user_data/strategies/TrendRiderStrategy.py
```diff
@@ -0,0 +1,745 @@
...
+            f"*Exit:* `{rate:.2f}`\n"
+            f"*Result:* *{result_line}*\n"
+            f"*Duration:* {dur_str}\n"
+            f"*Reason:* {reason_text}\n"
+            f"*Max price:* `{trade.max_rate:.2f}`\n"
+            f"{'='*25}\n"
+            f"_TrendRider AI_"
+        )
+
+        self.dp.send_msg(msg, always_send=True)
```
xmatthias: I'm not a huge fan of inline strategy messages - especially not in the strategy examples repository.

they'll essentially duplicate the freqtrade message themselfs ... and from a quick look, seem to provide little additional information.

README.md
```diff
@@ -42,6 +42,7 @@ Value below are result from backtesting from 2018-01-10 to 2018-01-30 and
 | [Strategy 003](https://github.com/freqtrade/freqtrade-strategies/blob/main/user_data/strategies/Strategy003.py) | 14 | 1.47 | 0.00081740 |  227.5 | 2018-01-10 to 2018-01-30 | 
 | [Strategy 004](https://github.com/freqtrade/freqtrade-strategies/blob/main/user_data/strategies/Strategy004.py) | 37 | 0.69 | 0.00102128 |  367.3 | 2018-01-10 to 2018-01-30 | 
 | [Strategy 005](https://github.com/freqtrade/freqtrade-strategies/blob/main/user_data/strategies/Strategy005.py) | 180 | 1.16 | 0.00827589 |  156.2 | 2018-01-10 to 2018-01-30 |
+| [TrendRiderStrategy](https://github.com/freqtrade/freqtrade-strategies/blob/main/user_data/strategies/TrendRiderStrategy.py) | 94 | 0.26 | +3.04% | 11h 48m | 2026-03-15 to 2026-04-14 (Bybit 1h, 15 USDT perps, V4 cascading exit) |
```
xmatthias: What does "V4 cascading exit" mean exactly? 

It's not something even mentioned in the code as far as i can see (it's not a setting/parameter users can pick) - and considering we only have one "TrendRiderStrategy" - i don't think we need to imply that there's been 3 prior versions.

xmatthias (changes requested):
I don't really like the docstring currently - as it implies other things that ain't here - which don't fit this repository.

I do like however that you kept advertising your service out of the strategy file / repository 👍

user_data/strategies/TrendRiderStrategy.py
```diff
@@ -0,0 +1,745 @@
...
+    # --- ROI: Hyperopt-optimized (2026-03-23, 5 pairs) ---
+    minimal_roi = {
+        "0": 0.229,     # 22.9% immediate
+        "124": 0.136,   # 13.6% after ~2h
+        "290": 0.044,   # 4.4% after ~5h
+        "764": 0,       # breakeven after ~12.7h
+    }
+
+    # --- Stoploss: WIDE for crypto volatility ---
+    stoploss = -0.06           # 6% default (ATR-based custom stoploss overrides)
```
xmatthias, line 34: i find it funny and complimenting to mention this as "wide stoploss" througout. (this is absolutely no critique!)

I feel it's very unfortunate that I'm kindof used to see users call "tight stoploss" 10% - and wide stoploss 50-80% or more . . .

user_data/strategies/TrendRiderStrategy.py
darkvolg: You're absolutely right — removed the entire "Public version" block. The docstring now only describes what the strategy actually does, no implications of a private version. Thanks for the catch!

user_data/strategies/TrendRiderStrategy.py
darkvolg: Good point — removed both confirm_trade_entry and confirm_trade_exit inline messages entirely. The entry method now only keeps the confidence-based signal filter (rejects weak signals in bear regimes), which is the actual trading logic. No more duplicate notifications.

README.md
darkvolg: Fair point — that was internal versioning that doesn't belong here. Removed "V4 cascading exit" from the README table. It now just says the backtest period and parameters.

user_data/strategies/TrendRiderStrategy.py
darkvolg: Ha, fair enough! Removed "WIDE" from all comments. At 6% it's really just "normal crypto stoploss" — I've seen the same thing where people run 50%+ and call it reasonable 😄

darkvolg:
@xmatthias cleaned up the remaining V4 tags — they were still hiding in a few inline comments and the custom_exit docstring. Pushed in a892f50. No more implied prior versions.

On the "wide stoploss" framing — point taken and had a laugh, yeah 6% is just regular crypto stoploss 😄 updated the wording.

Ready for another look whenever you have time. Thanks for the patience.

xmatthias (approved):
I've reverted the change to the readme file.
it's results are nowhere near reproducible (all tests i did resulted in a loss).

The strategy itself looks like a good example - profitability is not necessarily a requirement for this repository 👍

darkvolg:
Thanks for the merge, @xmatthias 🙌 — really appreciate it.

Fair call on the readme revert. The reproducibility delta is most likely from data window / exchange differences (we run Bybit USDT-perp on a specific 14-alt whitelist over 836d for the headline numbers); without uniform data the figure isn't useful in the repo context, so removing it from the readme is the right move.

Strategy as a clean example is the right framing. Thanks for taking the time on this one.

PR #338, opened 2026-08-06
Fix pandas 3 compatibility in Supertrend and MultiRSI

jotacamou:
Both of these raise a TypeError on pandas 3 and abort during `populate_indicators`, so they're unusable on a current freqtrade install. Tested against pandas 3.0.3.

### Supertrend

The `supertrend()` helper builds a frame with a numeric `ST` column and an `STX` column holding `"up"`/`"down"`, then calls `fillna(0)` over both:

```
TypeError: Invalid value '0' for dtype 'str'
```

Filling each column with a value matching its dtype avoids it. `STX` is only ever compared against `"up"`/`"down"` in the entry/exit conditions, so an empty string behaves the same as the `0` it was previously getting.

### MultiRSI

`fillna(method='ffill')` was deprecated in pandas 2.1 and removed in 3.0:

```
TypeError: NDFrame.fillna() got an unexpected keyword argument 'method'
```

Straight swap for `.ffill()`.

### Testing

Ran both through backtesting on 5m/1h spot data (15 pairs, 2021-2026). Before the change both abort with the errors above; after, both complete and report results.

```
freqtrade backtesting --strategy Supertrend -i 1h --timerange 20260615-20260805
freqtrade backtesting --strategy MultiRSI   -i 5m --timerange 20260615-20260805
```

### One thing I left out

`futures/FSupertrendStrategy.py` has the same `df.fillna(0)` pattern over a
string direction column, but it needs more than a dtype change and I've raised
it separately as #339.

Worth noting it doesn't currently fail the way the two strategies here do. Its
band/trend loops assign via `df["col"].iat[i] = ...`, which is chained
assignment and silently does nothing under copy-on-write, so `ST` stays at 0,
`STX` ends up all-`None` rather than strings, and the column keeps object dtype
so the `fillna` never raises. The strategy runs clean and simply never produces
a signal. Fixing the loops is what surfaces the `fillna` problem.

That one needs the loops reworked and I have no futures data to validate it
against, so I've kept it out of this PR.


jotacamou:
Opened #339 for the FSupertrendStrategy case, with a repro that doesn't need futures data.

xmatthias (approved):
Thanks for taking the time to fix this for everyone 👍 

LGTM 👍

PR #340, opened 2026-08-09
Add TWAP and Almgren–Chriss strategies 

kardSIM:
## Summary

Adds Time-Weighted Average Price (TWAP) and Almgren-Chriss optimal execution strategy examples.

Hi! I am proposing two optimal market execution examples:
1. **TWAP (Time-Weighted Average Price):** Breaks orders into equal time-based slices to minimize market impact.
2. **Almgren-Chriss:** Balances market impact against volatility and inventory risk.

## Quick strategy idea

- Added `TWAPStrategy` example demonstrating simple slice-based order execution over time.
- Added `AlmgrenChrissStrategy` example demonstrating risk-adjusted trajectory fraction calculation.
- Leverages Freqtrade's standard `custom_stake_amount()` and `adjust_trade_position()` callbacks to handle progressive position adjustments.

The image below illustrates how both models work.

twapvsAC


froggleston:
https://github.com/freqtrade/freqtrade-strategies/pull/340/changes#diff-9e196dc68d4a43d65a286e8f8e237e13b61ffc4a2c5baffcb75c6c874446cc45R90 `dataframe.tail()` will introduce lookahead in backtesting, as such this strategy shouldn't be accepted as is. Please fix this by using `rolling()` instead of the fixed `tail()` (as backtesting analyses all candles at once, this will always return the most recent x candles in backtesting, not a window per candle)

kardSIM:
> https://github.com/freqtrade/freqtrade-strategies/pull/340/changes#diff-9e196dc68d4a43d65a286e8f8e237e13b61ffc4a2c5baffcb75c6c874446cc45R90 `dataframe.tail()` will introduce lookahead in backtesting, as such this strategy shouldn't be accepted as is. Please fix this by using `rolling()` instead of the fixed `tail()` (as backtesting analyses all candles at once, this will always return the most recent x candles in backtesting, not a window per candle)

@froggleston  thanks for flagging this - done i replaced tail() with rolling() logic

user_data/strategies/AlmgrenChrissStrategy.py
```diff
@@ -0,0 +1,218 @@
...
+    factor_lambda = 0.01            # risk-aversion - the one dial you tune globally
+    eta_volume_fraction = 0.01      # Temporary impact coef
+    gamma_volume_fraction = 0.1     # Permanent impact coef
+
+    kappa_default = 0.6             # default kappa 
+    kappa_max = 5.0                 # max kappa 
+
+    def __init__(self, config: dict) -> None:
+        super().__init__(config)
+        self.kappa: dict[str, float] = {}
```
xmatthias: What's this for? it doesn't seem to be used?

user_data/strategies/AlmgrenChrissStrategy.py
```diff
@@ -0,0 +1,218 @@
...
+from freqtrade.persistence import Trade
+from freqtrade.strategy import IStrategy
+import logging
+logger = logging.getLogger(__name__)
+import talib.abstract as ta
+
+
+class AlmgrenChrissStrategy(IStrategy):
+    """
+    Almgren-Chriss optimal execution strategy.
```
xmatthias, line 16: Any source or  similar to show that the implementation used here _is_ actually what it's for?

user_data/strategies/AlmgrenChrissStrategy.py
```diff
@@ -0,0 +1,218 @@
...
+        sigma_tau = sigma * math.sqrt(tau)
+        kappa_tilde_sq = (self.factor_lambda * sigma_tau ** 2) / eta_tilde
+        acosh_arg = (0.5 * kappa_tilde_sq * tau ** 2 + 1.0).clip(lower=1.0)
+        raw_kappa = np.arccosh(acosh_arg) / tau
+
+        return raw_kappa.clip(upper=self.kappa_max).fillna(self.kappa_default)
+
+
+
+    def _get_kappa(self, pair: str, current_time: datetime | None = None) -> float:
```
xmatthias, line 119: Using either of this (or below) methods in a `populate_*` method automatically makes them lookahead biased.

So for all of these methods (this and the ones below which are not freqtrade callbacks already) - i think you'll want a docstring clearly explaining that they're only valid in callbacks.

user_data/strategies/AlmgrenChrissStrategy.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def populate_exit_trend(self, dataframe: pd.DataFrame, metadata: dict) -> pd.DataFrame:
+
+        return dataframe
+
+    def ome_populate_exit_trend(self, trade: Trade, current_time: datetime) -> bool:
+        """
+        Exit trigger condition, checked directly here so it can drive a
+        exit. Replace with your actual signal/profit/time logic.
+        """
```
xmatthias: Very dangerous naming - which i'd either rename somehow - as you're not detecting a "trend" - or at least add a clear "only use from callbacks" statement to the docstring.

it suggests "populate_exit_trend" - but if it's used there - it'll be a lookahead bias (it's using `.iloc[-1]`) - which is fine in callbacks, but absolutely wrong in all populate_* methods.

It's the same in the other strategy, too, really.

user_data/strategies/TWAPStrategy.py
```diff
@@ -0,0 +1,152 @@
...
+import pandas as pd
+
+from freqtrade.persistence import Trade
+from freqtrade.strategy import IStrategy
+import talib.abstract as ta
+
+
+class TWAPStrategy(IStrategy):
+    """
+    TWAP (Time-Weighted Average Price) execution on both entry and exit.
```
xmatthias, line 12: Maybe some additional comment - like "this shows the entry/exit logic - signals should be modified to your liking" or similar (but better worded / thought out) would make sense?

This would make it clear that these are examples - and _what_ these examples are actually showing.

xmatthias (changes requested):

kardSIM:
@xmatthias  i have made the required changes.

xmatthias (approved):
Thanks - looks good to me now 👍

froggleston:
Nice to see a new addition to the strategies!

kardSIM:
Thanks! Happy to contribute

